Issue #7, opened 2017-06-21
Fails to scan kops created clusters.


zapman449:
kops deploys most of the kubernetes binaries inside docker containers.  So the output looks like this:

```
$ ./kube-bench master
[INFO] Using config file: {A PATH}/config.yaml
[WARN] kube-apiserver: command not found on path - version check skipped
[WARN] kube-scheduler: command not found on path - version check skipped
[WARN] kube-controller-manager: command not found on path - version check skipped
[WARN] config file /etc/kubernetes/apiserver does not exist
[WARN] config file /etc/kubernetes/scheduler does not exist
[WARN] config file /etc/kubernetes/controller-manager does not exist
[WARN] config file /etc/kubernetes/config does not exist
[WARN] config file /etc/etcd/etcd.conf does not exist
[WARN] config file /etc/sysconfig/flanneld does not exist
yaml: line 61: did not find expected key
```

Please let me know if I can help with further triage

lizrice:
Thanks @zapman449! Just to be clear, is that {A PATH} a redaction, or literally what's output?

lizrice:
Sorry about this - I think you were the victim of a badly-formatted YAML file that got checked in!  Could you try again please, making sure you have the latest version by re-pulling aquasec/kube-bench:latest and reinstalling before you re-run it? 

I think you will still get warning messages about missing config files (we need to figure out how to handle them being installed to different places and with different names by different tools) but you should at least get some test output

zapman449:
Fair.  I tried again, but this time it's worse:

```
$ ./kube-bench master
wait: error running audit command exit status 1
```

lizrice:
Hi @zapman449 - we now have 
 * a flag for specifying the installation tool e.g. --installation kops 
 * the ability to specify in config.yaml the binaries and config files as used by different installation tools. 

However, we don't know whether we have the right config & binary locations for kops, so if you have a moment to give it a whirl that would be wonderful! 

There's also a --verbose flag for getting more information in the event of errors

zapman449:
Thank you. I was able to generate a report but with several warnings/caveats.

```
$ ./kube-bench master --installation kops
apiserver: command not found in path, error: exec: "apiserver": executable file not found in $PATH
scheduler: command not found in path, error: exec: "scheduler": executable file not found in $PATH
controller-manager: command not found in path, error: exec: "controller-manager": executable file not found in $PATH
[...]

PR #11, opened 2017-06-30
Resolve issue #7 wait: error running audit command exit status 1.

ttousai:
This is caused by a command in the audit pipeline (for example
ps -ef | grep kube-apiserver) failing. The causes of this failure
in my testing is usually a missing config file.

Extensive refactor and correction in verification code to check for
config files and binaries.

Replace joncalhoun/pipes with implementation using exec.Cmds so errors
are visible and can be handled when audit pipeline commands fail.

Change some audit commands
from: ps -ef | grep <cmd> | grep -v
to:   ps -C <something> -o comm,args --no-headers

which is simpler to work with.

### DIFF
--- a/cfg/federated.yaml
+++ b/cfg/federated.yaml
@@ -9,7 +9,7 @@ groups:
   checks:
      - id: 3.1.1
        text: "Ensure that the --anonymous-auth argument is set to false (Scored)"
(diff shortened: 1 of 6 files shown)


ttousai:
Left some debug commands in the code.

Issue #17, opened 2017-07-05
Add -installation <type> option to select between different installation types


lizrice:
A new -installation flag to choose the default config file and executable names for different installation types. 

For example, 
* `kube-bench master -installation kops` will look for config files and executables in the default locations installed by kops for a master node (see issue #7)
* `kube-bench node -installation hyperkube` looks for the defaults installed by hyperkube (see issue #10). 
* `kube-bench node -installation kubeadm` looks for the defaults installed by kubeadm. 

Using the flag means kube-bench picks from a set of default configurations to be defined in config.yaml. The user can override the defaults by modifying config.yaml.

lizrice:
[kubeadm](https://kubernetes.io/docs/admin/kubeadm/) creates these config files for a master node:
- /etc/kubernetes/admin.conf
- /etc/kubernetes/kubelet.conf
- /etc/kubernetes/controller-manager.conf
- /etc/kubernetes/scheduler.conf 

And it looks as though it's just /etc/kubernetes/kubelet.conf for a regular node. 

(see https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/)

Executables for kubeadm are as we currently have them in kube-bench

PR #18, opened 2017-07-10
Issues #17, #16

ttousai:
Support --installation option. Currently supports kops, hyperkube, kubeadm and the default installation. Will need some fleshing out for the bin/conf paths (#17).

Also added a --verbose flag to print verbose messages for debugging (#16).

### DIFF
--- a/check/check.go
+++ b/check/check.go
@@ -96,7 +96,6 @@ func (c *Check) Run(verbose bool) {
 	cs := c.Commands
 
 	// Initialize command pipeline
-	// cs[0].Stderr = os.Stderr
 	cs[n-1].Stdout = &out
 	i := 1
 
@@ -113,7 +112,6 @@ func (c *Check) Run(verbose bool) {
 			),
 		)
 
-		// 	cs[i].Stderr = os.Stderr
 		i++
 	}
 
@@ -150,7 +148,7 @@ func (c *Check) Run(verbose bool) {
 		i++
 	}
 
-	if verbose {
+	if verbose && errmsgs != "" {
 		fmt.Fprintf(os.Stderr, "%s\n", errmsgs)
 	}
 
--- a/check/check.go
+++ b/check/check.go
@@ -106,7 +106,7 @@ func (c *Check) Run(verbose bool) {
 		cs[i-1].Stdout, err = cs[i].StdinPipe()
 		errmsgs += handleError(
 			err,
-			fmt.Sprintf("check.Run: Audit %s failed\nfailing command: %s",
+			fmt.Sprintf("failed to run: %s\nfailed command: %s",
 				c.Audit,
 				cs[i].Args,
 			),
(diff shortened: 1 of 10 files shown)


cmd/common.go
```diff
@@ -129,34 +269,27 @@ func cleanIDs(list string) []string {
 // Any check failing here is a show stopper.
```
lizrice: Should have spotted this before but this comment isn't true

cmd/common.go
```diff
@@ -256,36 +389,31 @@ func verifyBin(binPath []string) []string {
...
-func verifyKubeVersion(binPath []string) []string {
+func verifyKubeVersion(b string) []string {
 	// These executables might not be on the user's path.
 	// TODO! Check the version number using kubectl, which is more likely to be on the path.
 	var w []string
 
-	for _, b := range binPath {
-		_, err := exec.LookPath(b)
-		if err != nil {
-			w = append(w, fmt.Sprintf("%s: command not found on path - version check skipped\n", b))
```
lizrice: Let's keep this warning as part of the output if the command isn't found on the path - this is useful information to have even if --verbose is not on.

cmd/common.go
```diff
@@ -256,36 +389,31 @@ func verifyBin(binPath []string) []string {
...
+	cmd := exec.Command(b, "--version")
+	out, err := cmd.Output()
+	errmsgs += handleError(
+		err,
+		fmt.Sprintf("verifyKubeVersion: failed\nCommand:%s", cmd.Args),
+	)
 
-		matched := strings.Contains(string(out), kubeVersion)
-		if !matched {
-			w = append(w, fmt.Sprintf(
```
lizrice: Any reason not to leave in the "expected" and "got" in the warning output? It seems potentially useful information

ttousai:
Nice! This is pretty detailed! I will review and update.

cmd/common.go
ttousai: I think I will have to spend sometime looking at error handling as a whole. For now I will restore this message but only on asserting the verbose flag.

check/check.go
```diff
@@ -88,38 +96,50 @@ func (c *Check) Run() {
...
 	// Start command pipeline
 	i = 0
 	for i < n {
 		err := cs[i].Start()
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "%s: %s\n", cs[i].Args, err)
-		}
+		errmsgs += handleError(
+			err,
+			fmt.Sprintf("check.Run: Audit %s failed\nfailing command: %s",
```
ttousai: I try to capture 2 pieces of info with that error message:
1. The entire audit pipeline that failed.
2. Which specific command in that pipeline failed.

So something like:
`
Failed to run: commandline
Failing command: actual command
`

cmd/common.go
ttousai: Missed it too!

cmd/common.go
ttousai: I run into issues where the error message looks like;
```
expected: 1.6, got: 
```
This usually happens when the kubernetes binary I'm using for version check is not running. So I tried to solve but removing this error message and replacing it with something more generic.

PR #28, opened 2017-07-25
Improve error handling.

ttousai:
Check out what I have been working on for error handling:

1. added support for glog log levels: 
    $ kube-bench master
    $ kube-bench master -v 1
    $ kube-bench master -v 2  # can get very loud
1. Error handling is now at call site, no longer propagating errors.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -17,7 +17,6 @@ package cmd
 import (
 	"fmt"
 	"io/ioutil"
-	"os"
 	"strings"
 
 	"github.com/aquasecurity/kube-bench/check"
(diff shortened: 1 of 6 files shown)


cmd/util.go
```diff
@@ -20,11 +21,35 @@ var (
...
 
-func handleError(err error, context string) (errmsg string) {
+func printWarn(msg string) {
+	fmt.Fprintf(os.Stderr, "[%s] %s\n",
+		colors[check.WARN].Sprintf("%s", check.WARN),
+		msg,
+	)
+}
+
+func printfWarn(msg string) string {
```
lizrice: This is a small thing but I think for readability it would be better to call this function `sprintWarn`, to be clearer about the fact this is returning a string (that would be more consistent with the difference between printf and sprintf)

cmd/common.go
```diff
@@ -120,63 +117,50 @@ func runChecks(t check.NodeType) {
...
 }
 
 // verifyNodeType checks the executables and config files are as expected
 // for the specified tests (master, node or federated).
-func verifyNodeType(t check.NodeType) []string {
-	var w []string
-	// Always clear out error messages.
-	errmsgs = ""
-
+func verifyNodeType(t check.NodeType) {
```
lizrice: I haven't had a chance to try this yet so just want to check - will this still work if the output is in JSON format? In the JSON case we want to only emit the test results (so that it's machine readable). It might be fine, if the warnings are going to stderr and the JSON output is going to stdout.

ttousai: I checked I think it will work for JSON. All the error messages are printed to stderr.

cmd/util.go
ttousai: I agree.

cmd/common.go
lizrice: Finally found time to test it and I agree, it's sending messages to the right place

ttousai:
PR updated. I will recommend we apply PR#19 first.

PR #34, opened 2017-08-11
Use kubectl to check the kubernetes version

### LINKED ISSUE #15: Version checks using kubectl
It seems to be very common that the user won't have all the kubernetes executables in their path (e.g. kube-apiserver) but that they can still query the running version using kubectl, so this would be a better approach.

lizrice:
Fixes #15

### DIFF
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"regexp"
 	"strings"
 
 	"github.com/aquasecurity/kube-bench/check"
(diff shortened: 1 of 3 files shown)


cmd/util.go
```diff
@@ -123,25 +124,61 @@ func verifyBin(binPath ...string) {
...
-	_, err := exec.LookPath(b)
+	_, err := exec.LookPath("kubectl")
 	if err != nil {
 		continueWithError(err, sprintlnWarn("Kubernetes version check skipped"))
 		return
 	}
 
-	cmd := exec.Command(b, "--version")
+	cmd := exec.Command("kubectl", "version")
 	out, err := cmd.Output()
```
ttousai: @lizrice Can we make kubectl version command's output only be printed when log level is set to 1? I think it clutters the output of kube-bench.

PR #36, opened 2017-08-12
Issue 25

ttousai:
This issue was caused by shell builtins failing the binary lookup test quietly and returns WARN. To fix the config file tests are rewritten to:
 /bin/sh -c 'if test -e /etc/foo; then stat ... ; fi'

### DIFF
--- a/check/check.go
+++ b/check/check.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"regexp"
 	"strings"
 )
 
@@ -75,8 +76,7 @@ func (c *Check) Run(verbose bool) {
 
 	// Check if command exists or exit with WARN.
 	for _, cmd := range c.Commands {
-		_, err := exec.LookPath(cmd.Path)
-		if err != nil {
+		if !isShellCommand(cmd.Path) {
 			c.State = WARN
 			return
 		}
@@ -111,7 +111,6 @@ func (c *Check) Run(verbose bool) {
 				cs[i].Args,
 			),
 		)
-
 		i++
 	}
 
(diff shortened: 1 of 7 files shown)


check/check.go
```diff
@@ -166,22 +166,63 @@ func (c *Check) Run() {
...
+		// CHALLENGE: The input text may contain quoted strings that
+		// must be passed as a unit to exec.Cmd.
+		// eg. bash -c 'foo bar'
+		// 'foo bar' must be passed as unit to exec.Cmd if not the command
+		// will fail when it is executed.
+		// eg. exec.Cmd("bash", "-c", "foo bar")
+		//
+		// PROBLEM: Current solution assumes the grouped string will always
+		// be at the end of the input text.
+		re := regexp.MustCompile(`^(.*)(['"].*['"])$`)
```
lizrice: Not sure I understand the comment here - if you drop the $ at the end of the regex won't it catch the text inside quotes, even if it's not at the end of the input text?

check/check.go
```diff
@@ -166,22 +166,63 @@ func (c *Check) Run() {
...
 
 		cmd := exec.Command(cs[0], cs[1:]...)
 		cmds = append(cmds, cmd)
 	}
 
 	return cmds
 }
+
+func isShellCommand(s string) bool {
+	cmd := exec.Command("/bin/sh", "-c", "command -v "+s)
```
lizrice: Would it make sense to check whether something is not a builtin by using exec.LookPath?

check/check.go
```diff
@@ -166,22 +166,63 @@ func (c *Check) Run() {
...
 	cmds := []*exec.Cmd{}
 
 	cp := strings.Split(s, "|")
-	// fmt.Println("check.toCommand:", cp)
 
 	for _, v := range cp {
 		v = strings.Trim(v, " ")
-		cs := strings.Split(v, " ")
+
+		// TODO:
```
ttousai: I will agree that it is done. Only not a robust solution.

check/check.go
ttousai: Yes it will. But I'm not really confident in the regexp, especially if there are multiple quoted strings.

check/check.go
ttousai: No I don't see anything in the LookPath documentation that indicates it does that.

check/test_test.go
```diff
@@ -30,79 +32,66 @@ func init() {
...
 		panic("Failed reading test data: " + err.Error())
 	}
-	controls, err = NewControls(MASTER, in)
+
+	// substitute variables in data file
+	user := os.Getenv("USER")
+	s := strings.Replace(string(in), "$user", user, -1)
+
+	controls, err = NewControls(MASTER, []byte(s))
+	// controls, err = NewControls(MASTER, in)
```
ttousai: Yes we can.

PR #45, opened 2017-08-30
Some better file locations for kubeadm and hyperkube

lizrice:


lizrice:
This is superceded by another PR I'm about to make

PR #46, opened 2017-08-30
Add initial support for multiple versions of kubernetes.

ttousai:


ttousai:
@lizrice I have attempted to clean up as much as possible.

cmd/federated.go
```diff
@@ -33,7 +33,7 @@ func init() {
 	federatedCmd.PersistentFlags().StringVarP(&federatedFile,
 		"file",
 		"f",
-		cfgDir+"/federated.yaml",
+		"/federated.yaml",
```
lizrice: I haven't checked this, but it seems surprising that it works? Looks like it's saying that the default for a federated file is ./federated.yaml? 

In fact, I'm not sure we even need this ability to override the file name. We could keep things simpler by omitting these flags (for master, node & federated file locations) - wdyt? We could take those out under a separate issue though if you agree.

cmd/util.go
```diff
@@ -90,65 +90,39 @@ func verifyBin(bin string, psFunc func(string) string) bool {
...
-
-	msg := checkVersion("Client", string(out), major, minor)
-	if msg != "" {
-		continueWithError(fmt.Errorf(msg), msg)
-	}
-
-	msg = checkVersion("Server", string(out), major, minor)
-	if msg != "" {
-		continueWithError(fmt.Errorf(msg), msg)
+		return nil
```
lizrice: The test for len(out) == 0 allowed for the case where kubectl gives client version but not server.  We get an error but there is still have some kubectl output to check so we can give the user some info about what version is running. So I think we should add it back in.

check/controls.go
```diff
@@ -23,10 +23,11 @@ import (
 
 // Controls holds all controls to check for master nodes.
 type Controls struct {
-	ID     string `yaml:"id"`
-	Text   string
-	Type   NodeType
-	Groups []*Group
+	ID      string `yaml:"id"`
+	Version string
```
lizrice: Do we use this Version field anywhere? I couldn't see it but perhaps I missed it.

ttousai:
@lizrice I will like to close this request. I have a cleaner PR that I will like to submit. Some of the questions asked here will apply there. I attempted to merge the latest HEAD of master branch. But I think it will complicate the PR so I will submit this PR without the merge.

ttousai:
I will close this PR and note [PR#52](https://github.com/aquasecurity/kube-bench/pull/52) supersedes this.

Issue #49, opened 2017-09-12
The controls for master - admission control showing wrong status


junaid18183:
For master , controls 1.1.11,1.1.12,1.1.13,1.1.14,1.1.15,1.1.25,1.1.28,1.1.33  showing wrong status.

example - 

```
 # ./kube-bench master -c 1.1.11,1.1.12,1.1.13,1.1.14,1.1.15,1.1.25,1.1.28,1.1.33
Unexpected Client version 1.6
Unexpected Server version 1.6
[INFO] 1 Master Node Security Configuration
[INFO] 1.1 API Server
[PASS] 1.1.11 Ensure that the admission control policy is not set to AlwaysAdmit (Scored)
[FAIL] 1.1.12 Ensure that the admission control policy is set to AlwaysPullImages (Scored)
[FAIL] 1.1.13 Ensure that the admission control policy is set to DenyEscalatingExec (Scored)
[FAIL] 1.1.14 Ensure that the admission control policy is set to SecurityContextDeny (Scored)
[PASS] 1.1.15 Ensure that the admission control policy is set to NamespaceLifecycle (Scored)
[FAIL] 1.1.25 Ensure that the admission control policy is set to PodSecurityPolicy (Scored)
[FAIL] 1.1.28 Ensure that the admission control policy is set to ServiceAccount (Scored)
[FAIL] 1.1.33 Ensure that the admission control policy is set to NodeRestriction (Scored)

== Remediations ==
1.1.12 Edit the /etc/kubernetes/apiserver file on the master node and set the KUBE_ADMISSION_CONTROL parameter to "--admission-control=...,AlwaysPullImages,..."
1.1.13 Edit the /etc/kubernetes/apiserver file on the master node and set the KUBE_ADMISSION_CONTROL parameter to "--admission-control=...,DenyEscalatingExec,..."
1.1.14 Edit the /etc/kubernetes/apiserver file on the master node and set the KUBE_ADMISSION_CONTROL parameter to "--admission-control=...,SecurityContextDeny,..."
1.1.25 Follow the documentation and create Pod Security Policy objects as per your environment. Then, edit the /etc/kubernetes/apiserver file on the master node and set the KUBE_ADMISSION_CONTROL parameter to "--admission-control=...,PodSecurityPolicy,..."
[...]

PR #50, opened 2017-09-13
Fixed issue admission control showing wrong status

junaid18183:
Fixed issue of The controls for master - admission control showing wrong status #49
Changed the regex


lizrice:
Terrific, thank you!

PR #52, opened 2017-09-18
Support multiple kubernetes version

ttousai:
This PR is supposed to be a cleaned up version of [PR#46](https://github.com/aquasecurity/kube-bench/pull/46).

Some questions on that PR apply here.

To keep this clean I did not merge kube-bench HEAD.

### DIFF
--- a/cfg/1.6/federated.yaml
+++ b/cfg/1.6/federated.yaml
@@ -1,0 +1,285 @@
+---
+controls:
+version: 1.6
+id: 3
+text: "Federated Deployments"
+type: "federated"
+groups:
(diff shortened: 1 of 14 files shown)


lizrice:
This is looking good. However I ran it on a machine with 1.6 and got the following: 

Unexpected Client version 1.6
Unexpected Server version 1.6

From an extremely quick look, I think this is because we always verify against kubeMajorVersion and kubeMinorVersion which are currently hard-coded to 1.7?

lizrice:
I've opened #53 to cover the errors running on 1.6

Issue #56, opened 2017-09-29
Add v1.8 support


luxas:
Kubernetes v1.8 is now released :tada:, let's get some YAML profiles for this fancy new version as well :smile:

lizrice:
Yes! There will be an update to the CIS spec very soon and then we can generate the YAML to match 😀

luxas:
Cool! @lizrice Would you mind joining a SIG Cluster Lifecycle video call btw to discuss how we can have a better kubeadm <-> kube-bench integration? I'd eventually like to use sth like this in our automated tests.
(separate issue though)

lizrice:
I'd love to. This is Tuesdays 5pm UK time, right? I can't make it this week but could do the 10th, if that works.

luxas:
Yes, sounds good!

timothysc:
@lizrice I'd be happy to help review e2e-prs that you might want to add.  Please denote any additions as [Feature:xyz] to allow for selective enablement. 

/cc @mattmoyer

bitvector2:
What is the current status on this?  We would very much like to use and can help with the work effort...

luxas:
@lizrice ^

@bitvector2 On updating this project for v1.8 or integrating with kubeadm?

lizrice:
1.8 support is now released (yay!) 

I'll close this issue now and open another one for kubeadm integration. We would very much welcome help and/or pointers on how to get started with that (I'm not familiar with the codebase over there).

PR #59, opened 2017-10-15
Add kubernetes 1.8 support

ttousai:
Implement #56

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -68,6 +68,7 @@ func runChecks(t check.NodeType) {
 	binmap := getBinaries(typeConf)
 	confmap := getConfigFiles(typeConf)
 	podspecmap := getPodSpecFiles(typeConf)
+	unitfilemap := getUnitFiles(typeConf)
 
 	switch t {
 	case check.MASTER:
@@ -90,6 +91,7 @@ func runChecks(t check.NodeType) {
 	s = makeSubstitutions(s, "bin", binmap)
 	s = makeSubstitutions(s, "conf", confmap)
 	s = makeSubstitutions(s, "podspec", podspecmap)
+	s = makeSubstitutions(s, "unitfile", unitfilemap)
 
 	glog.V(1).Info(fmt.Sprintf("Using config file: %s\n", viper.ConfigFileUsed()))
 	glog.V(1).Info(fmt.Sprintf("Using benchmark file: %s\n", path))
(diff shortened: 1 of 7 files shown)


cmd/common.go
```diff
@@ -67,6 +67,8 @@ func runChecks(t check.NodeType) {
 	// checks that the executables we need for the node type are running.
 	binmap := getBinaries(typeConf)
 	confmap := getConfigFiles(typeConf)
+	podspecmap := getPodSpecFiles(typeConf)
```
lizrice: Is there are reason to split out pod spec and unit files from the configuration files? Maybe there is a good reason to do this but it seems like it makes it more confusing for the user trying to edit the config.yml.  

Also, although the latest version of the benchmark is the first to talk about using these pod spec files I think kubernetes was actually using this approach from at 1.7 (maybe earlier?). So I don't think it's a bad thing if people run the 1.7 or 1.6 tests and look for the config in these locations. Or am I mistaken?

cmd/util.go
```diff
@@ -147,6 +147,68 @@ func getConfigFiles(v *viper.Viper) map[string]string {
 	return confmap
 }
 
+// getPodSpecFiles finds which of the set of candidate podspec files exist
+func getPodSpecFiles(v *viper.Viper) map[string]string {
```
lizrice: Even if there's a reason to split out pod and unit files, why not just add a name parameter to the config file function and use it for all three cases? I think that's the only real difference between these three get*Files() functions

cmd/common.go
ttousai: If we don't separate podspec from config files anyone running our latest kube-bench on k8s v1.7 or older will have the remediation applying to the wrong file because of what we will get for $apiserverconf.

```
Kubernetes v1.8
Remediation:
Edit the API server pod specification file $apiserverconf on the master node and set the below parameter.
--anonymous-auth=false
```

```
Kubernetes v1.7
Remediation:
Edit the $apiserverconf file on the master node and set 
the KUBE_ALLOW_PRIV parameter to \"--allow-privileged=false\"
```
But separating podspec allows me to say $apiserverpodspec for the v1.8 remediation.

cmd/util.go
ttousai: Yes. I agree with that.

lizrice: OK, great, could you make that change please and then I think we'll be good to go!

cmd/common.go
lizrice: OK, that makes sense, thanks

lizrice: OK, that makes sense, thanks

PR #61, opened 2017-10-31
added saving results to PostgreSQL DB as a JSONB document

bitvector2:
majority of changes in savePgsql() in common.go

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -18,8 +18,13 @@ import (
 	"fmt"
 	"io/ioutil"
 
+	"os"
+	"time"
+
 	"github.com/aquasecurity/kube-bench/check"
 	"github.com/golang/glog"
+	"github.com/jinzhu/gorm"
+	_ "github.com/jinzhu/gorm/dialects/postgres"
 	"github.com/spf13/viper"
 )
 
@@ -118,7 +123,17 @@ func runChecks(t check.NodeType) {
 
 		fmt.Println(string(out))
 	} else {
-		prettyPrint(controls, summary)
+		// if we want to store in PostgreSQL, convert to JSON and save it
+		if (summary.Fail > 0 || summary.Warn > 0 || summary.Pass > 0) && pgSql {
+			out, err := controls.JSON()
+			if err != nil {
+				exitWithError(fmt.Errorf("failed to output in JSON format: %v", err))
+			}
+
+			savePgsql(string(out))
+		} else {
+			prettyPrint(controls, summary)
+		}
 	}
 }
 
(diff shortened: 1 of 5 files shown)


cmd/common.go
```diff
@@ -104,7 +109,17 @@ func runChecks(t check.NodeType) {
...
 	} else {
-		prettyPrint(controls, summary)
+		// if we want to store in PostgreSQL, convert to JSON and save it
+		if (summary.Fail > 0 || summary.Warn > 0 || summary.Pass > 0) && pgSql {
+			out, err := controls.JSON()
+			if err != nil {
+				exitWithError(fmt.Errorf("failed to output in JSON format: %v", err))
+			}
+
+			savePgsql(string(out))
```
lizrice: On successfully writing to the database, would it be worth outputting some message to let the user know that it happened?

cmd/common.go
```diff
@@ -154,3 +169,51 @@ func prettyPrint(r *check.Controls, summary check.Summary) {
...
+		"PGSQL_USER":     viper.GetString("PGSQL_USER"),
+		"PGSQL_DBNAME":   viper.GetString("PGSQL_DBNAME"),
+		"PGSQL_SSLMODE":  viper.GetString("PGSQL_SSLMODE"),
+		"PGSQL_PASSWORD": viper.GetString("PGSQL_PASSWORD"),
+	}
+
+	for k, v := range envVars {
+		if v == "" {
+			err := fmt.Errorf("Environment variable %s missing", envVarsPrefix+"_"+k)
+			panic(err)
```
lizrice: Please could you use the exitWithError() function rather than panic()? (Here and a couple more instances in the same function.)

cmd/common.go
```diff
@@ -154,3 +169,51 @@ func prettyPrint(r *check.Controls, summary check.Summary) {
 		summary.Pass, summary.Fail, summary.Warn,
 	)
 }
+
+func savePgsql(jsonInfo	 string) {
```
lizrice: if you don't mind I think this is big enough to warrant starting a new file called, say, postgres.go or database.go. 

(I don't really like that this file is called common.go either, but that's another matter!)

cmd/common.go
bitvector2: yep

cmd/common.go
bitvector2: yep

cmd/common.go
bitvector2: yep

bitvector2:
Should be good now..

Issue #66, opened 2017-11-02
Unnecessary warnings about missing config files


lizrice:
Running on 1.8 we expect to find pod spec YAML files for many components. However, the code assumes they will have config files so we get a number of unnecessary warnings e.g. 

[ec2-user@k8s344-vm0 ~]$ ./kube-bench master
[WARN] Missing config file for flanneld
[WARN] Missing config file for kubernetes
[WARN] Missing config file for apiserver
[WARN] Missing config file for scheduler
[WARN] Missing config file for controllermanager
[WARN] Missing config file for etcd
[WARN] Missing config file for flanneld
[WARN] Missing config file for kubernetes

lizrice:
Fixed by #68

PR #68, opened 2017-11-03
Remove unnecessary warnings about missing config files.

ttousai:
issue #66


ttousai:
@lizrice I see. What do you think about having a default config `cfg/config` and a version specific config `cfg/1.6/config` to hold configurations which a specific to a kubernetes version.

We can make configuration settings in the latter override anything in the default after loading, this way we can specify which version need configfiles and which needs podspecs.

lizrice:
Yes, I think that would make sense

Issue #76, opened 2017-12-06
Container installation failing


jpweber:
When running the container installation command `docker run --rm -v `pwd`:/host aquasec/kube-bench:latest` it fails to play the kube-bench binary on the host. See output below.  

```
ubuntu@ip-10-1-1-215:~$ sudo docker run --rm -v `pwd`:/host aquasec/kube-bench:latest
cp: can't stat './kube-bench/cfg/*': Not a directory
cp: can't stat './kube-bench/kube-bench': Not a directory
===============================================
kube-bench is now installed on your host
Run ./kube-bench to perform a security check
===============================================
ubuntu@ip-10-1-1-215:~$ ls
cfg
ubuntu@ip-10-1-1-215:~$
```

this is image id dd0785160c94 
digest: sha256:896995e04785b95fe45de7e6207f68b7f207a461d169e27db8c8ebad2f2632d0

lizrice:
Fixed by #77

PR #77, opened 2017-12-11
Fixed paths in entrypoint.sh

jpweber:
The paths in the copy commands for cfg dir and kube-bench binary were not pointing to where those actually lived in the container.

Fix for issue #76


lizrice:
How embarrassing! Thanks so much for catching this @jpweber 

Fixes #76

PR #83, opened 2018-01-11
Add goreleaser configuration

jaxxstorm:
This adds goreleaser configuration and also updates the travis config for this project so it'll autobuild when news tags are pushed.

You'll need to add a token to your travis config for the org that has push permissions, and then set that as a secret environment variable `GITHUB_TOKEN` within travis.

Let me know if I can help in any way.

I omitted the homebrew configuration for now, as aquasecurity doesn't have a homebrew-tap repo. Let me know if you want additional config for that.


.goreleaser.yml
```diff
@@ -0,0 +1,23 @@
...
+      - linux
+    goarch:
+      - amd64
+# Archive customization
+archive:
+  format: tar.gz
+dockers:
+  -
+    binary: kube-bench
+    image: aquasecurity/kube-bench
```
lizrice: This is referring to our image on Docker Hub, right? The repo name is aquasec/kube-bench (aquasec not aquasecurity)

lizrice:
Thanks for this @jaxxstorm! 

@jerbia could you help with the travis token for this please?

.goreleaser.yml
jaxxstorm: Fixed!

jerbia:
Sure, I can help with the Travis build.

@jaxxstorm - can you send me more details on the Travis token? Where should I put it?

jerbia:
@jaxxstorm I've added the GITHUB_TOKEN to the .travis file.
Let me know if this is OK now.

jaxxstorm:
@jerbia @lizrice I've rebased this so there's no longer any merge conflicts. Would probably be a good idea to do a test tag and test release, but you should be good to try this out now

PR #90, opened 2018-01-24
Improve etcd data directory extraction

jhvhs:
- If data-dir is not the last argument, the remaining arguments
  are captured preventing the correct checking.

Signed-off-by: Konstantin Semenov <ksemenov@pivotal.io>


jhvhs:
@ttousai In fact, sed seems to implement a [very small subset of regex](https://www.gnu.org/software/sed/manual/html_node/Regular-Expressions.html). It's even hard to say what is `(\S*)` exactly matching. Looks like the correct expression would actually be
`sed 's%.*data-dir[= ]\([^ ]*\).*%\1%'`

lizrice:
Thanks @jhvhs!

PR #96, opened 2018-02-23
Update README.md to reflect that the --installation option has been r…

clemensw:
…emoved.


lizrice:
Thanks for spotting & fixing this @clemensw!

Issue #97, opened 2018-02-23
kube-bench assumes kubectl is available on node


clemensw:
kube-bench determines the kubernetes version via kubectl.  This isn't available on our kops installed node instances.  kubelet --version would return the node version. Alternatively kube-bench should gain an option to manually set the version.

lizrice:
Good idea - getting the version from the kubelet would be better. Tagging @lirkog86 who I think might be interested in this too?

PR #100, opened 2018-03-23
Use kubelet

philalex:
Try to use kubelet when kubectl is unavailable
Resole #97


philalex:
Done. Thanks for the review.

PR #108, opened 2018-04-13
Allow kubernetes version and config directory to be specified (resolves #107)

wmedlar:
Adds in the `--config-dir` and `--version` flags, allowing a user to specify the config directory containing benchmarks, and the target Kubernetes version, respectively.

PR #112, opened 2018-04-23
Support for Client Version and Server Version output.

oranmoshai:


oranmoshai:
It is related to the warning "[WARN] Unable to get kubectl version, using default version: 1.6" when running ./kube-bench node on aws eks node.

lizrice:
It really has to be the server version we get, not the client - and it's not uncommon to find mismatching client & servers installed. Perhaps EKS gives a different format output for `kubectl --version` for the server that doesn't match the current regex? If so, please open an issue showing what the output looks like, and then let's create a regex to match that.  

Alternatively, there are a couple of recent changes that might help instead: 

* #100  for getting the server version directly from kubelet (which has be present - kubectl doesn't have to be)  
* #108 adds flags so you can specify the version directly

PR #115, opened 2018-05-06
Allow kube-bench to be run from inside its container

### LINKED ISSUE #109: ps usage is not portable
[The current usage of `ps` to detect binaries](https://github.com/aquasecurity/kube-bench/blob/ef6c017f54520acfb9211a0ae6827ce2b8f5a82d/cmd/util.go#L78) is not portable. [Busybox's ps](https://busybox.net/downloads/BusyBox.html) (scroll way down) is [POSIX-compatible](https://github.com/mirror/busybox/blob/master/procps/ps.posix), although the options are undocumented in the command help. Unfortunately `-C`, `-o cmd`, and `--no-headers` aren't part of the POSIX spec; rather they're are GNU-specific options, I think.

```shell
$ docker run --rm gcr.io/google-containers/kube-apiserver:v1.9.3 ps -C test
ps: invalid option -- C
BusyBox v1.28.0 (2018-01-16 23:29:21 UTC) multi-call binary.

Usage: ps [-o COL1,COL2=HEADER] [-T]

Show list of processes

	-o COL1,COL2=HEADER	Select columns for display
	-T			Show threads
```

Doesn't look like they're included in the [BSD](https://www.freebsd.org/cgi/man.cgi?ps(1))/[OSX](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/ps.1.html) ps either, but that's a much smaller concern. (`-C` is related to CPU percentage calculation, not process name)

```shell
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.13.1
BuildVersion:	17B48
$ man ps
[...]
SYNOPSIS
     ps [-AaCcEefhjlMmrSTvwXx] [-O fmt | -o fmt] [-G gid[,gid...]] [-g grp[,grp...]] [-u uid[,uid...]] [-p pid[,pid...]] [-t tty[,tty...]] [-U user[,user...]]
     ps [-L]
[...]
```

[github.com/shirou/gopsutil](https://github.com/shirou/gopsutil) may be useful as a portable solution.

### LINKED ISSUE #114: kube-bench should be runnable without installation
I want to use kube-bench to score my new Kubernetes cluster but I don't want to install software on each of my servers. If we are already delivering a Docker image, can kube-bench just run from there instead of needing to be copied to each host?

Seems like if you just wanted to score against the standard CIS benchmark, you shouldn't have to install anything and we should be able to publish kubectl commands to just execute kube-bench against the cluster and display the results.

wmedlar:
This PR adds the following capabilities to the kube-bench docker image:

- build `kube-bench` binary from local source (rather than remote with `go get`); this allows for easier testing of local development
- verify dependency versions with `glide`
- disables building with cgo to run in alpine
- installs `procps` to allow for GNU-specific `ps` flags while running the in a container (partially resolves #109)
- adds `install` command to image entrypoint to retain the ability to install from a container
- modifies default container behavior to execute `kube-bench $ARGS` (resolves #114)


lizrice:
Thank you for the PR, it's neat to be able to run this from inside a container :-) 

I am concerned about changing the default behaviour as there are tools built on top of kube-bench that may rely on this. @jerbia @nuwaida @bgeesaman any concerns? 

If we're going to add dependency management I would rather go with dep (as I'm under the impression that glide is being deprecated). In any case the glide files are not here?  Perhaps dependency management would be better as the subject of a separate PR.

Dockerfile
```diff
@@ -1,13 +1,21 @@
...
+ADD cmd/ cmd/
+RUN CGO_ENABLED=0 go install -a -ldflags '-w'
 
-FROM alpine:latest
-WORKDIR /
-COPY --from=0 /go/bin/kube-bench /kube-bench 
-COPY --from=0 /go/src/github.com/aquasecurity/kube-bench/cfg /cfg
-COPY --from=0 /go/src/github.com/aquasecurity/kube-bench/entrypoint.sh /entrypoint.sh
-ENTRYPOINT /entrypoint.sh
+FROM alpine:latest AS run
```
lizrice: I'd rather we specify a version for repeatable(ish) builds

bgeesaman:
No concerns from me as the Sonobuoy plugin is not currently under active development.  Thanks for thinking of me, though!

wmedlar:
Agreed, migration to `dep` would make more sense as a separate PR; I can handle that one too if the maintainers don't have bandwidth.

Installation as the default behavior can be added with a simple `CMD ["install"]` at the end of the dockerfile, I just didn't think to do it at the time, though I'd have to wonder how many people are using the docker-based installation as part of their automation. It might really make more sense to have go-releaser build binaries and upload them to the repo release page.

Is that @bgeesaman of the Hacking and Hardening Kubernetes talk? You gave one of the best presentations at KubeCon Austin! I was linking people to your slides for weeks! 😄

bgeesaman:
Hah, yep.  Thank you for the kind words! 👍

jerbia:
@lizrice no concerns from my side.

Issue #120, opened 2018-05-11
Add tests to Travis to confirm that config files and executable are in place


lizrice:
We've had a few issues where the installation has failed after a change - would be good to have a test in the pipeline to make sure that the Docker image has the files in the expected places

PR #122, opened 2018-05-11
Fail fast in container entrypoint

wmedlar:
This sets `-e` to exit early if any line in the entrypoint script fails. Should help catch installation issues once Travis tests are set up for the Docker images (#120).


lizrice:
Thanks!

PR #123, opened 2018-05-11
Add tip about the `--version` flag to error output

Tenzer:
If people are trying to use the Docker image to check their cluster, there's a
big likelyhood of them hitting the error message saying that either `kubectl`
or `kubelet` need to be found in order for `kube-bench` to be able to determine
the Kubernetes version in use.

This adds a tip that the version can be specified manually with the `--version`
flag which is a lot easier than having to make a new Docker image with the
right version of `kubelet`/`kubectl` in order for `kube-bench` to work.


lizrice:
Good idea, thanks!

PR #124, opened 2018-05-11
Adding a test install to travis job

lizrice:
Checks that cfg directory and kube-bench directories are in place on the host after running the install command from within a container


lizrice:
Fixes #120

PR #130, opened 2018-05-15
A bunch of text edits

ttousai:


cmd/root.go
```diff
@@ -46,7 +46,7 @@ var (
 var RootCmd = &cobra.Command{
 	Use:   os.Args[0],
 	Short: "Run CIS Benchmarks checks against a Kubernetes deployment",
-	Long:  `This tool runs the CIS Kubernetes 1.6 Benchmark v1.0.0 checks.`,
+	Long:  `This tool runs the CIS Kubernetes Benchmark (http://www.cisecurity.org/benchmark/kubernetes/)`,
 }
```
lizrice: Good idea to make this more future-proof!

PR #131, opened 2018-05-15
Fix booleans comparaison issue125

philalex:
Do case insensitive comparaison for booleans. 
Code could be rewritten like:
````go
value := strings.ToLower(flagVal)
if value != "false" && value != "true" {
          value = flagVal
}
result = value == t.Compare.Value
````
But i don't like this style ...


ttousai:
Thanks @philalex 

LGTM

PR #134, opened 2018-05-28
fix grep string for check 1.4.11 and 1.4.12

hutr:
check 1.4.11 and 1.4.22 shows FAIL even when permissions is correct.

### DIFF
--- a/cfg/1.8/master.yaml
+++ b/cfg/1.8/master.yaml
@@ -942,8 +942,9 @@ groups:
 
   - id: 1.4.11
     text: "Ensure that the etcd data directory permissions are set to 700 or more restrictive (Scored)"
-    audit: ps -ef | grep $etcdbin | grep -- --data-dir | grep -v grep | sed 's%.*data-dir[= ]\([^ ]*\).*%\1%' | xargs stat -c %a
-    test_items:
+    audit: ps -ef | grep $etcdbin | grep -- --data-dir | sed 's%.*data-dir[= ]\([^ ]*\).*%\1%' | xargs stat -c %a
+    tests:
+      test_items:
       - flag: "700"
         compare:
           op: eq
@@ -959,7 +960,7 @@ groups:
 
   - id: 1.4.12
     text: "Ensure that the etcd data directory ownership is set to etcd:etcd (Scored)"
-    audit: ps -ef | grep $etcdbin | grep -- --data-dir | grep -v grep | sed 's%.*data-dir[= ]\(\S*\)%\1%' | xargs stat -c %U:%G
+    audit: ps -ef | grep $etcdbin | grep -- --data-dir | sed 's%.*data-dir[= ]\(\S*\)%\1%' | xargs stat -c %U:%G
     tests:
       test_items:
       - flag: "etcd:etcd"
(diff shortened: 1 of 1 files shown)


cfg/1.8/master.yaml
```diff
@@ -942,9 +942,8 @@ groups:
 
   - id: 1.4.11
     text: "Ensure that the etcd data directory permissions are set to 700 or more restrictive (Scored)"
-    audit: ps -ef | grep $etcdbin | grep -v grep | sed 's%.*data-dir[= ]\([^ ]*\).*%\1%' | xargs stat -c %a
-    tests:
-      test_items:
+    audit: ps -ef | grep $etcdbin | grep -- --data-dir | grep -v grep | sed 's%.*data-dir[= ]\([^ ]*\).*%\1%' | xargs stat -c %a
+    test_items:
```
lizrice: Looks like this is missing `tests:`?

lizrice: I think this is missing `tests:`

cfg/1.8/master.yaml
```diff
@@ -942,9 +942,8 @@ groups:
 
   - id: 1.4.11
     text: "Ensure that the etcd data directory permissions are set to 700 or more restrictive (Scored)"
-    audit: ps -ef | grep $etcdbin | grep -v grep | sed 's%.*data-dir[= ]\([^ ]*\).*%\1%' | xargs stat -c %a
-    tests:
-      test_items:
+    audit: ps -ef | grep $etcdbin | grep -- --data-dir | grep -v grep | sed 's%.*data-dir[= ]\([^ ]*\).*%\1%' | xargs stat -c %a
```
lizrice: I don't think there is any need for the grep -v grep step now that we are checking for --data-dir

cfg/1.8/master.yaml
```diff
@@ -960,7 +959,7 @@ groups:
 
   - id: 1.4.12
     text: "Ensure that the etcd data directory ownership is set to etcd:etcd (Scored)"
-    audit: ps -ef | grep $etcdbin | grep -v grep | sed 's%.*data-dir[= ]\(\S*\)%\1%' | xargs stat -c %U:%G
+    audit: ps -ef | grep $etcdbin | grep -- --data-dir | grep -v grep | sed 's%.*data-dir[= ]\(\S*\)%\1%' | xargs stat -c %U:%G
```
lizrice: I think grep -v grep  is now unnecessary here too

cfg/1.8/master.yaml
hutr: Added tests:

cfg/1.8/master.yaml
hutr: You are correct about this. Removed grep -v grep

cfg/1.8/master.yaml
hutr: Removed grep -v grep

Thanks for correction!

hutr:
Thanks for the correction.

Added tests: for 1.4.11
Removed grep -v grep for 1.4.11 and 1.4.12
Also fixed sed string for 1.4.12 to fetch right --data-dir folder

PR #138, opened 2018-06-15
Rule node 2.2.4 is not correct

jgsqware:


cfg/1.8/node.yaml
```diff
@@ -368,8 +368,7 @@ groups:
       scored: true
 
     - id: 2.2.4
-      text: "Ensure that the kubelet service file permissions are set to 644 or
-      more restrictive (Scored)"
+      text: "Ensure that the kubelet.conf file ownership is set to root:root (Scored)"
```
ttousai: @jgsqware thank you for catching this.

However I checked again the CIS Kubernetes Benchmark v1.2.0 document for kubernetes 1.8.0 Node benchmark 2.2.4 and it is:
```
2.2.4 Ensure that the kubelet service file ownership is set to root:root
(Scored)
```
Can you please change this in your PR?

jgsqware:
Yep done thanks

lizrice:
Thank you @jgsqware!

Issue #126, opened 2018-05-14
Files ending in .yaml seem to now end in .manifest


micah:
In 1.9, I'm told things like this:

```1.1.1 Edit the API server pod specification file /etc/kubernetes/manifests/kube-apiserver.yaml on the master node and set the below parameter.```

but it seems like the files all end with .manifest instead of .yaml, perhaps this is a 1.9 change?

lizrice:
Hi @micah, which tool you're using to install Kubernetes? That might affect what the files are called and where they are installed to. 

You can simply modify the config files yourself to suit your system, but I'm interested to know what installer you're using in case it's something we should consider providing an option for.

micah:
> Hi @micah, which tool you're using to install Kubernetes? That might
> affect what the files are called and where they are installed to.

I used kubespray. I was unsure why they were called this, and thought
perhaps it was a 1.9 change.

> You can simply modify the config files yourself to suit your system,
> but I'm interested to know what installer you're using in case it's
> something we should consider providing an option for.

I'm guessing it might be something worth considering, since kubespray is
an incubator project, and a fairly popular mechanism for installing!

-- 
        micah

lizrice:
Closing as the .manifest files are now included in the config file. Please re-open if you still see a problem

PR #140, opened 2018-06-29
Inlcude .manifest extension config files for kops & kubespray

lizrice:
Adds support for kops & kubespray configs that use the .manifest extension instead of .yaml


lizrice:
Should fix #126

Issue #136, opened 2018-06-13
Getting Error While Running Kube-Bench on AWS from local machine 


abhinavwwefan:
I have a cluster configured in the AWS environment. Running the Kubernetes Infrastructure 

Now i am connecting to pods and nods through the kubectl using the kube-conf file 

I have installed Kube-bench on my local machine and trying to execute it on master and nodes but getting the below response.

> abhinav:kube-bench abhinav$ ./kube-bench master
> No CIS spec for 1.10 - using tests from CIS 1.2.0 spec for Kubernetes 1.8
> 
> need apiserver executable but none of the candidates are running
> abhinav:kube-bench abhinav$ ./kube-bench node
> No CIS spec for 1.10 - using tests from CIS 1.2.0 spec for Kubernetes 1.8
> 
> need kubelet executable but none of the candidates are running

Can you help me how i can run the kube-bench on the infrastructure from my local machine 
or if i have to setup with a Kubernetes infra or on my EC2 instance

ttousai:
Hello @abhinavwwefan kube-bench must be run on your infrastructure. You can use the following to run on your infrastructure:

Run the master check
```
kubectl run --rm -i -t kube-bench-master --image=aquasec/kube-bench:latest --restart=Never --overrides="{ \"apiVersion\": \"v1\", \"spec\": { \"hostPID\": true, \"nodeSelector\": { \"kubernetes.io/role\": \"master\" }, \"tolerations\": [ { \"key\": \"node-role.kubernetes.io/master\", \"operator\": \"Exists\", \"effect\": \"NoSchedule\" } ] } }" -- master --version 1.8
```

Run the node check
```
kubectl run --rm -i -t kube-bench-node --image=aquasec/kube-bench:latest --restart=Never --overrides="{ \"apiVersion\": \"v1\", \"spec\": { \"hostPID\": true } }" -- node --version 1.8
```

lizrice:
There isn't a set of tests defined by the CIS for 1.10 and at the moment we require you to specify the version (with the `--version 1.8` flag). 

@ttousai I think it would be better if we automatically defaulted to 1.8 tests if the version number is higher (for now - there should be a new CIS spec coming out soon for 1.11). Wdyt?

ttousai:
@lizrice I think that will work. What do you think about making it the latest kube-bench supported CIS version.

lizrice:
[...]

PR #141, opened 2018-06-29
Default version 

lizrice:
As per discussion in #136. 

Oops, I have left the config file changes in this branch for the .manifest extension support. Never mind, hopefully it will merge OK.


lizrice:
Fix for discussion under #136

Issue #142, opened 2018-07-02
Error running this on GKE 1.8.10 clusters


sauryadas:
Version check failed: need kubectl or kubelet binaries to get kubernetes version.
Alternately, you can specify the version with --version
pod default/kube-bench-node terminated (Error)

lizrice:
The error message suggests that neither kubectl nor kubelet are in the current path. Can you tell us a bit more about how you ran this (directly as a binary by invoking kube-bench, or with a kubectl command?) and where you ran it?

sauryadas:
kubectl run --rm -i -t kube-bench-node --image=aquasec/kube-bench:latest --restart=Never --overrides="{ \"apiVersion\": \"v1\", \"spec\": { \"hostPID\": true } }" -- node --version 1.8

alkar:
I've had a similar issue on a 1.10 cluster in `kops` - fetching the proper `kubectl` in the container fixes it.

fatal-exception:
thanks @alkar , are you able to give any more detail about how you did the fix? I get the same issue with our 1.9 cluster built using kops

hairyhenderson:
I'm also seeing this on a kops-provisioned cluster with Kubernetes v1.9.8:

```
$ kubectl run --rm -i -t kube-bench-master --image=aquasec/kube-bench:latest --restart=Never --overrides="{ \"apiVersion\": \"v1\", \"spec\": { \"hostPID\": true, \"nodeSelector\": { \"kubernetes.io/role\": \"master\" }, \"tolerations\": [ { \"key\": \"node-role.kubernetes.io/master\", \"operator\": \"Exists\", \"effect\": \"NoSchedule\" } ] } }" -- master --version 1.8 -v 2

Version check failed: need kubectl or kubelet binaries to get kubernetes version.
Alternately, you can specify the version with --version
pod default/kube-bench-master terminated (Error)
```

subnova:
The aquasec/kube-bench:fix-typo image does not suffer from the problem, guess introduced by #141

hairyhenderson:
yeah, and I can confirm that the `lizrice-patch-1` tag also does not suffer from the issue

stafot:
@subnova on EKS cluster also the `aquasec/kube-bench:fix-typo` does the trick. 
So command for nodes that works is the following and not the one on `README.md` 
[...]

PR #147, opened 2018-07-26
Shouldn't need kubelet or kubectl if version specified

lizrice:
Fix for #142


lizrice:
I"m going to go ahead and merge this as it's a real problem. @ttousai please still shout if you have any comments on the change.

PR #157, opened 2018-08-18
Migrate to bench common

ttousai:


check/test.go
```diff
@@ -1,194 +0,0 @@
...
-		var flagVal string
-		isset := match
-
-		if isset && t.Compare.Op != "" {
-			// Expects flags in the form;
-			// --flag=somevalue
-			// --flag
-			// somevalue
-			//pttn := `(` + t.Flag + `)(=)*([^\s,]*) *`
-			pttn := `(` + t.Flag + `)(=)*([^\s]*) *`
```
lizrice: Need to check why we removed this comma in the regexp, as the version in bench_common still has the version with the comma

check/controls_test.go
```diff
@@ -1,41 +0,0 @@
...
-	"io/ioutil"
-	"testing"
-
-	yaml "gopkg.in/yaml.v2"
-)
-
-const cfgDir = "../cfg/"
-
-// validate that the files we're shipping are valid YAML
-func TestYamlFiles(t *testing.T) {
```
lizrice: Let's not lose this valuable test please (it runs in the automated build and prevents us from shipping a faulty test file)

check/test_test.go
```diff
@@ -1,121 +0,0 @@
...
-	"strings"
-	"testing"
-)
-
-var (
-	in       []byte
-	controls *Controls
-)
-
-func init() {
```
lizrice: Also let's make sure we keep these tests, they are useful

check/data
```diff
@@ -1,160 +0,0 @@
----
```
lizrice: See comment elsewhere, I want to keep the tests that use this data

check/controls_test.go
ttousai: @lizrice the plan is to move all the test to bench-common. But I see your point about testing test files. I'll add the TestYamlFiles back in.

check/test_test.go
ttousai: See previous comments.

check/data
ttousai: See previous comment on test files.

ttousai:
@lizrice I have added tests for the YAML definition files like I mentioned in my previous comments.

check/controls_test.go
lizrice: Thanks

check/test_test.go
lizrice: I see we have something in bench_common called TestTestExecute, but I think we are losing the set of actual test cases below, for specific kube-bench tests. Don't we need those?

check/data
lizrice: There are some new tests in bench_common that might cover at least some of this, but I suspect we are losing some useful tests here?

ttousai:
@lizrice I have pushed the tests which are at risk of getting lost to bench-common.

lizrice:
@ttousai I think we should close this PR as this set of changes is pretty out of date now. You OK with that?

ttousai:
@lizrice please go ahead.

PR #162, opened 2018-10-05
update to remove keep-terminated-pod-volumes

Sharma-Rajat:
This has been deprecated in k8s 1.8.

https://github.com/kubernetes/kubernetes/pull/47539

### DIFF
--- a/cfg/1.8/node.yaml
+++ b/cfg/1.8/node.yaml
@@ -159,25 +159,6 @@ groups:
     scored: true
 
   - id: 2.1.9
-    text: "Ensure that the --keep-terminated-pod-volumes argument is set to false (Scored)"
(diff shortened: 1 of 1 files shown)


lizrice:
Hi @sharma-rajat, thanks for your PR but unfortunately we're not going to accept it. kube-bench implements the tests as per the CIS spec, without being opinionated about that spec. 

There has just recently been a new version of the CIS spec published, and @ttousai is working on a new set of config files to implement those tests. This particular test has been retired in the new spec version so it won't be present in the new set of test config files. 

One of the benefits of the config file approach is that you can modify the tests to suit your own purposes locally. You might want to leave the numbering for the remaining tests intact so that it continues to match the spec

Sharma-Rajat:
@lizrice Thank you!! I didn't know this :)

PR #169, opened 2018-10-25
Fixing 1.3.7 on 1.11 master.

mikekim:
With multiple test items operator defaults to "and". In case of 1.3.7
the tests check whether --address flag is either set to 127.0.0.1 or not
set at all. Those conditions cannot be met at the same time.


lizrice:
@ttousai please could you review?

ttousai:
@mikekim thanks for catching this one.
@lizrice lgtm

PR #173, opened 2018-11-06
Fix check 1.1.37.

ttousai:
Check 1.1.37 should pass if --features-gate is not set or if features-gate is set to AdvancedAuditing=false.


ttousai:
@lizrice you are right about the bin_op. I have pushed the change.

lizrice:
Yes, I think the CIS describes the audit but our test doesn't fully implement it without the check on the --audit-policy-file so it would be good if you could add that as well (can be in a new PR though)

PR #180, opened 2018-11-16
Support kube-controller{-manager} as well

mstemm:
In some cases, the ps -C output for long processes like
kube-controller-manager may be truncated, meaning the search should be
for the truncated prefix.

This occurs with GKE and k8s 1.11 on minikube.

This fixes https://sysdig.atlassian.net/browse/SSPROD-1253.

### DIFF
--- a/cfg/config.yaml
+++ b/cfg/config.yaml
@@ -43,6 +43,7 @@ master:
   controllermanager:
     bins:
       - "kube-controller-manager"
+      - "kube-controller"
       - "hyperkube controller-manager"
       - "controller-manager"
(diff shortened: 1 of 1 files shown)


mstemm:
Sorry for closing this. I think there may be a wider set of necessary changes, including checking for prefixes for the api server as well. Also I wanted to take a look at the usage of `ps -C` to see if there was a better way of looking for the base process without being subject to prefix truncation.

ttousai:
@mstemm that will be awesome! Looking forward to your contribution.

PR #181, opened 2018-11-20
read config files from host /etc

lizrice:
I don't see how kube-bench can check the permissions on files unless it has access to them on the host, so I think we need to be mounting the /etc directory from the host. 

@ttousai can you double-check this please?

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,16 +25,16 @@ You can choose to
 
 ### Running inside a container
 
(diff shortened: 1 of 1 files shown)


ttousai:
@lizrice LGTM.

Another possible mount will be `etcd` data directory and WAL directories which are also checked and will be mounted on the host (checks 1.4.11 and 1.4.12).

lizrice:
Thanks @ttousai. I'll add mounting /var as well.

PR #200, opened 2019-01-14
warn osx limitation

spuder:


ttousai:
Thank you @spuder for pointing this out.

@lizrice I think we should remove the *installing from container* section altogether, there are quite a lot of ways to run kube-bench all ready.

lizrice:
This is a good point, thanks @spuder. I realised it's also true for Windows so I've added that into this PR. 

@ttousai the thing I like about this approach is that by installing and running directly on the host there's no danger of any issues related to not having the right access to the host config files or processes (which shouldn't be an issue, but we know from #198 it has been from time to time). So I'd rather keep it in there. (I suppose you could argue that you can always clone the repo onto the host, but the install approach means you don't need to have git installed - the less software installed on the host the better from an attack surface point of view).

PR #205, opened 2019-01-23
Fix 1.1.23 to check only false value for the flag

### LINKED ISSUE #204: Flags set vs. default values
I guess this is more of a question or starter for a discussion. :)

I've hit into one specific case:
```
1.1.23 Ensure that the --service-account-lookup argument is set to true (Scored)
```

Looking up the documentation on it at https://v1-11.docs.kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/
> --service-account-lookup     Default: true
>     If true, validate ServiceAccount tokens exist in etcd as part of authentication.

The test for it clearly verifies that the option is set to `true`:
```
- id: 1.1.23
    text: "Ensure that the --service-account-lookup argument is set to true (Scored)"
    audit: "ps -ef | grep $apiserverbin | grep -v grep"
    tests:
      test_items:
      - flag: "--service-account-lookup"
        compare:
          op: eq
          value: true
        set: true
```

So in this case, and there's few more cases like this, it's kinda false-positive IMO as the option defaults to true. Wouldn't it be better to write the test as checking that the option is not `false`, maybe something like this:
```
- id: 1.1.23
    text: "Ensure that the --service-account-lookup argument is set to true (Scored)"
    audit: "ps -ef | grep $apiserverbin | grep -v grep"
    tests:
      test_items:
      - flag: "--service-account-lookup"
        compare:
          op: neq # Dunno if there's neq operator in the framework :)
          value: false
```

So to make kube-bench quiet, we have to set many of the flags "unnecessarily" even though they do default to sensible values.

If there's been this sort of discussion already before and I was too blind to find it, please close this one. :) (and maybe point me to the previous results)

jnummelin:
From docs:
> --service-account-lookup     Default: true
> If true, validate ServiceAccount tokens exist in etcd as part of authentication.

So it does not really make sense to always check that the flag is set to `true`. Instead I flipped the test to check that it's NOT `false`.

Fixes #204

### DIFF
--- a/cfg/1.11/master.yaml
+++ b/cfg/1.11/master.yaml
@@ -366,15 +366,14 @@ groups:
     scored: true
 
   - id: 1.1.23
-    text: "Ensure that the --service-account-lookup argument is set to true (Scored)"
+    text: "Ensure that the --service-account-lookup argument is not set to false (Scored)"
     audit: "ps -ef | grep $apiserverbin | grep -v grep"
     tests:
       test_items:
       - flag: "--service-account-lookup"
         compare:
-          op: eq
-          value: true
-        set: true
+          op: noteq
+          value: false
     remediation: |
       Edit the API server pod specification file $apiserverconf
       on the master node and set the below parameter.


cfg/1.11/master.yaml
```diff
@@ -366,15 +366,14 @@ groups:
     scored: true
 
   - id: 1.1.23
-    text: "Ensure that the --service-account-lookup argument is set to true (Scored)"
+    text: "Ensure that the --service-account-lookup argument is not set to false (Scored)"
```
ttousai: @jnummelin please keep the original `text` for this check for compatibility with the CIS benchmark.

lizrice:
In the 1.4.0 version of the CIS spec this test has been corrected to say "Verify that if the --service-account-lookup argument exists it is set to true", but this has slipped through incorrectly in our 1.13 config files. I now think we should leave the 1.11 file as it is so that it reflects the CIS spec, but correct the 1.13 one. @ttousai would you mind doing that please?

yoavrotems:
We corrected to 1.13 one  in https://github.com/aquasecurity/kube-bench/pull/247 pull

jnummelin:
>  I now think we should leave the 1.11 file as it is so that it reflects the CIS spec

Kinda makes sense as kube-bench should follow the CIS spec.

I'll close this then.

PR #208, opened 2019-01-24
adding kubelet config.yaml path which is used in kubeadm for centos a…

OrlinVasilev:
…nd ubuntu

### DIFF
--- a/cfg/config.yaml
+++ b/cfg/config.yaml
@@ -80,7 +80,8 @@ node:
       - "kubelet"
     confs:
       - /etc/kubernetes/kubelet.conf
-      - /etc/kubernetes/kubelet 
+      - /etc/kubernetes/kubelet
+      - /var/lib/kubelet/config.yaml
     defaultconf: "/etc/kubernetes/kubelet.conf"
 
   proxy:


lizrice:
Oh... I think you have uncovered something more significant!

* The remediation text in 2.1.* tests refer to a "kubelet service file" in which you can change parameters. We use the substitution parameter $kubeletconf for this. Looking at the files on a kubeadm installation, this should be `/var/lib/kubelet/config.yaml` (as suggested in your PR here, @OrlinVasilev, thank you) 

* The tests in 2.2.* refer to a number of kubelet-related files. 2.2.9 and 2.2.10 are about `/var/lib/kubelet/config.yaml`, but we don't currently have a substitution parameter for that one.

* 2.2.1 and 2.2.2 refer to the kubelet.conf file `/etc/kubernetes/kubelet.conf`. This is the *kubeconfig file* for the kubelet - essentially the credentials for the kubelet, not its configuration parameters. 

* 2.2.3 and 2.2.4 refer to the service file, as in, a systemd service file. I need to check on this but I think this approach is now replaced by the `/var/lib/kubelet/config.yaml`

* 2.2.5-2.2.8 refer to other files that I think are OK. 

If the above is correct, I think we are currently using $kubeletconf to represent both the parameters file which should be `/var/lib/kubelet/config.yaml`, and the kubeconfig file at `/etc/kubernetes/kubelet.conf`. This is wrong, and we need another substitution parameter!

I'd suggest 
* make $kubeletconf represent the parameter config file: change the definition (and the default definition under `defaultconf`) to `/var/lib/kubelet/config.yaml` 
* have a new substitution parameter $kubeletkubeconfig (wdyt?) for the kubeconfig file at `/etc/kubernetes/kubelet.conf`- @ttousai will this need a change to the Go code to pick it up? 
* change 2.2.1 and 2.2.2 to use $kubeletkubeconfig

Wdyt? I would appreciate someone else checking my working here @OrlinVasilev @ttousai

lizrice:
@OrlinVasilev we think this is now sorted out with #228 - does this work for you?

OrlinVasilev:
@lizrice you looks great thanks

PR #211, opened 2019-01-29
Type skip and not scored checks

yoavAqua:
Support new check type - skip:
If a check is marked with type "skip", it will be marked as Info.

Support scored property:
If a check is not scored and is not marked with type skip, it will be marked as Warn.

### DIFF
--- a/check/check.go
+++ b/check/check.go
@@ -71,13 +71,20 @@ type Check struct {
 	TestInfo    []string    `json:"test_info"`
 	State       `json:"status"`
 	ActualValue string `json:"actual_value"`
+	Scored      bool   `json:"scored"`
 }
 
 // Run executes the audit commands specified in a check and outputs
 // the results.
 func (c *Check) Run() {
-	// If check type is manual, force result to WARN.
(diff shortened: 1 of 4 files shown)


ttousai:
@yoavAqua @lizrice  I have similar code in https://github.com/aquasecurity/kube-bench/tree/ocp-configs branch where I keep all ocp config changes.

If this change is not blocking anything I will recommend we close this PR in favor of my changes and merge as unit when the ocp config work is done, to keep things simple and avoid a conflict.

@lizrice WDYT?

lizrice:
This is a good change, it matches what we have in bench-common, and it even has a test. @ttousai I think if you concentrate on just the OCP YAML config that will be easy to merge.

ttousai:
@lizrice just to be clear. 

I already had skip logic code in the ocp-config branch (https://github.com/aquasecurity/kube-bench/commit/6d76965e50f6ccc3342c5aa48183149dc096e7e5). 

We will have to keep this in mind in later when we merge my changes in that branch.

PR #214, opened 2019-02-05
Add mount our own configuration

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/4818

It allows us to run directly kube bench with our own setup. I have tried to mount a configmap so image can be the same one but it fails parsing YAML (I tried to fix but I dont want to expend too much time), so now it uses our own image


pipo02mix:
Bad target sorry

PR #218, opened 2019-02-17
Bugfix: Logging warning instead of printing

### LINKED ISSUE #217: Printing errors although json flag is set
When the flag --json is set, the only output to be printed should be the data in json format.
There are 2 errors that are printed without considering the use of that flag which makes automations to fail on parsing the output as json.

yoavAqua:
Made all the warnings to be logged and not printed, so when using the json flag the output will be only in json format.

fix #217

### DIFF
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -33,20 +33,6 @@ func init() {
 	statFunc = os.Stat
 }
 
-func printlnWarn(msg string) {
-	fmt.Fprintf(os.Stderr, "[%s] %s\n",
-		colors[check.WARN].Sprintf("%s", check.WARN),
-		msg,
-	)
-}
-
-func sprintlnWarn(msg string) string {
-	return fmt.Sprintf("[%s] %s",
-		colors[check.WARN].Sprintf("%s", check.WARN),
-		msg,
-	)
-}
-
 func exitWithError(err error) {
 	fmt.Fprintf(os.Stderr, "\n%v\n", err)
 	os.Exit(1)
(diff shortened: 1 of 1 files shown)


cmd/util.go
```diff
@@ -33,13 +33,6 @@ func init() {
...
 }
 
-func printlnWarn(msg string) {
-	fmt.Fprintf(os.Stderr, "[%s] %s\n",
-		colors[check.WARN].Sprintf("%s", check.WARN),
-		msg,
-	)
-}
-
 func sprintlnWarn(msg string) string {
```
lizrice: I think we could delete `sprintlnWarn` this function too now?

cmd/util.go
```diff
@@ -336,7 +329,7 @@ func getVersionFromKubectlOutput(s string) string {
 	serverVersionRe := regexp.MustCompile(`Server Version: v(\d+.\d+)`)
 	subs := serverVersionRe.FindStringSubmatch(s)
 	if len(subs) < 2 {
-		printlnWarn(fmt.Sprintf("Unable to get kubectl version, using default version: %s", defaultKubeVersion))
+		glog.V(1).Info(fmt.Sprintf("Unable to get kubelet version, using default version: %s", defaultKubeVersion))
```
lizrice: Have you checked this goes to stderr? If output is in JSON format going to stdout, it shouldn't have any messages interspersed in it.

cmd/util.go
yoavAqua: It's not relevant to this issue but yes, this function is not called by anyone so I'm deleting it.

cmd/util.go
yoavAqua: These errors will only be printed if --v 1 is specified, checked it

PR #219, opened 2019-02-18
run the kubelet binary to get the kubernetes version on pks environment

nshauli:
On PKS platform kubelet is not in the path, so kube-bench fails to get the version and doesn't scan the host. The suggested solution was to run the kubelet binary, assuming it will always be found at /var/vcap/data/packages/kubernetes/*/bin/kubelet.
The fix was tested on PKS platform and it worked fine.

### DIFF
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -303,6 +303,12 @@ func getKubeVersion() (string, error) {
 	if err != nil {
 		_, err = exec.LookPath("kubelet")
 		if err != nil {
+			// It might be a PKS platform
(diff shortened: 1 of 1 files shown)


nshauli:
The wildcard replaces only the node id of the node the kube-bench is running on. Each node has it's own kubelet binary, so actually instead of running /var/vcap/data/packages/kubernetes/NODE_ID/bin/kubelet, I made it more generic with the wildcard and so there is no need in such case to run 'find' command to find the binary and then run it.

nshauli:
This pull request fixes the issue for PKS only, so I created new pull request which fixes the issue for all platforms (https://github.com/aquasecurity/kube-bench/pull/222)

Issue #220, opened 2019-02-19
Instructions for running `kube-bench` on AWS EKS


complex64:
We are looking to run `kube-bench` against an EKS cluster in AWS. This is EKS platform `eks.1` at Kubernetes version `1.11`, i.e. everything is up-to-date.

We have been taking the approach of running the job YAMLs included in this repository. However with no success. I'll now try to document the issues as well as possible, and would be grateful for your help in resolving this.

--- 

## 1. The EKS control plane is not user accessible

EKS, as a managed service offered by AWS, does not permit access to the control plane, that is master nodes, to schedule workloads.

The node selecor and tolerations in [job-master.yaml](https://raw.githubusercontent.com/aquasecurity/kube-bench/master/job-master.yaml) have no effect. The pod will not be scheduled anywhere and the job's pod will be left in state "Pending" indefinitely. Example pod events for a pod `kube-bench-master-wlmxl`:

```
  Type     Reason             Age                 From                Message
  ----     ------             ----                ----                -------
  Warning  FailedScheduling   11s (x25 over 35s)  default-scheduler   0/9 nodes are available: 9 node(s) didn't match node selector.
```
What approach would you recommend instead for security benchmarks of AWS EKS master nodes?

## 2. The Node Job has $PATH issues on AWS EKS

We are adjusting the command in [job-node.yaml](https://raw.githubusercontent.com/aquasecurity/kube-bench/master/job-node.yaml) for `-v3` prior running.

The job's pods will cycle quickly from Pending, ContainerCreating, to Error. Inspection of the logs yields:

```
I0219 11:26:28.944088   20794 util.go:132] Looking for config for version 1.11
I0219 11:26:28.944134   20794 util.go:137] Looking for config file: cfg/1.11/node.yaml
I0219 11:26:28.944175   20794 common.go:65] Using benchmark file: cfg/1.11/node.yaml
I0219 11:26:28.944408   20794 common.go:77] Using config file: cfg/1.11/config.yaml
I0219 11:26:28.944632   20794 util.go:57] [ps -C hyperkube -o cmd --no-headers]: fork/exec /usr/bin/ps: no such file or directory
I0219 11:26:28.944662   20794 util.go:284] executable 'hyperkube kubelet' not running
I0219 11:26:28.944819   20794 util.go:57] [ps -C kubelet -o cmd --no-headers]: fork/exec /usr/bin/ps: no such file or directory
[...]

PR #221, opened 2019-02-19
Set PATH for exec of /bin/ps

complex64:
@lizrice, here you go! :)

Fixes https://github.com/aquasecurity/kube-bench/issues/220

### DIFF
--- a/job-node.yaml
+++ b/job-node.yaml
@@ -10,6 +10,9 @@ spec:
       - name: kube-bench
         image: aquasec/kube-bench:latest
         command: ["kube-bench","node"]
+        env:
+          - name: PATH
+            value: /bin:/usr/bin:/usr/local/bin
         volumeMounts:
         - name: var-lib-kubelet
           mountPath: /var/lib/kubelet


job-node.yaml
```diff
@@ -10,6 +10,9 @@ spec:
       - name: kube-bench
         image: aquasec/kube-bench:latest
         command: ["kube-bench","node"]
+        env:
+          - name: PATH
+            value: /bin:/usr/bin:/usr/local/bin
```
lizrice: On further reflection, I'm wondering if the modified PATH shouldn't include the locations that are currently defined in the image: 
```
$ docker run --rm -it --entrypoint=/bin/sh aquasec/kube-bench
/opt/kube-bench # echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

Wdyt?

Pluies: @lizrice I've been investigating this issue as it's been causing us trouble as well. We're on EKS using the AWS-provided image as well.

I think the issue comes from the fact that the container mounts `/usr/bin` from the host into `/usr/bin` within kube-bench, and a `ps` happens to be available in there. Well, _available_... Kind of:
```
/opt/kube-bench # ls -lah /usr/bin/ps
-rwxr-xr-x    1 root     root       97.8K Aug  1  2018 /usr/bin/ps
/opt/kube-bench # /usr/bin/ps
/bin/sh: /usr/bin/ps: not found
```
I think that by virtue of existing (at least to `ls`) and being earlier in the path, this `ps` gets picked up before the "real" `ps` from the container at `/bin/ps`:
```
/opt/kube-bench # ps
  PID TTY          TIME CMD
20719 pts/0    00:00:00 sh
23531 pts/0    00:00:00 ps
/opt/kube-bench # /bin/ps
  PID TTY          TIME CMD
20719 pts/0    00:00:00 sh
23535 pts/0    00:00:00 ps
```
But because it's a program from the host, some protection mechanism kicks in and prevents kube-bench from executing it, hence the error message from #220:
```
I0219 11:26:28.944632   20794 util.go:57] [ps -C hyperkube -o cmd --no-headers]: fork/exec /usr/bin/ps: no such file or directory
I0219 11:26:28.944662   20794 util.go:284] executable 'hyperkube kubelet' not running
```

I'm not sure what the best solution is here, to be honest. Potentially just rearrange the path to keep all the existing entries, but have `/bin` be a higher priority than `/usr/bin`? This works for me for example:
```yaml
env:
- name: PATH
  value: /usr/local/sbin:/usr/local/bin:/usr/sbin:/sbin:/bin:/usr/bin
```

lizrice: I like the look of @Pluies solution here 
```
env:
- name: PATH
  value: /usr/local/sbin:/usr/local/bin:/usr/sbin:/sbin:/bin:/usr/bin

complex64:
Is this still relevant with https://github.com/aquasecurity/kube-bench/pull/260 merged, or can be closed?

Issue #209, opened 2019-01-25
Add configuration for kubernetes version 1.13


aspartame21:

lizrice:
We'll add a new set of configs when the next release of the CIS benchmark comes out. The next release (1.4, which has tests for Kubernetes 1.13) is at Draft stage right now.

PR #226, opened 2019-02-27
add new config files from the new CIS Kubernetes Benchmark

yoavrotems:
there is a new update at CIS_Kubernetes_Benchmark_v1.4.0 for Kubernetes 1.13


cfg/1.13/master.yaml
```diff
@@ -0,0 +1,1500 @@
+---
+controls:
+version: 1.11
```
lizrice: Change to 1.13 please

cfg/1.13/node.yaml
```diff
@@ -0,0 +1,480 @@
+---
+controls:
+version: 1.11
```
lizrice: And here too

ttousai:
@lizrice @yoavrotems please also consider the changes made in #228.

lizrice:
That's a very good point. cfg/1.13/node.yaml will need to take the same changes that are going into cfg/1.11/node.yaml under #228

lizrice:
This will fix #209

PR #228, opened 2019-02-27
Fix issues with checks for kubelet configuration files

ttousai:
This is a more detailed solution to the issue reported in the PR #208. It fixes issues with variable substitutions for node benchmarks.

There are 3 major config files for kubelets k8s nodes:
- /etc/kubernetes/kubelet.conf ($kubeletkubeconfig)
- /var/lib/kubelet.config.yaml ($kubeletconf)
- /etc/systemd/system/kubelet.service.d/10-kubeadm.conf ($kubeletsvc)

In the previous versions of kube-bench there was no distinction between these files and results of running node checks were misleading. This PR adds the distinction and also adds support for a new type of variable `kubeconfig`.

### DIFF
--- a/cfg/1.8/config.yaml
+++ b/cfg/1.8/config.yaml
@@ -9,36 +9,13 @@
 
 master:
   apiserver:
-    confs:
-      - /etc/kubernetes/manifests/kube-apiserver.yaml
-      - /etc/kubernetes/manifests/kube-apiserver.manifest
     defaultconf: /etc/kubernetes/manifests/kube-apiserver.yaml
 
   scheduler:
-    confs: 
-      - /etc/kubernetes/manifests/kube-scheduler.yaml
-      - /etc/kubernetes/manifests/kube-scheduler.manifest
     defaultconf: /etc/kubernetes/manifests/kube-scheduler.yaml
 
   controllermanager:
-    confs:
-      - /etc/kubernetes/manifests/kube-controller-manager.yaml
-      - /etc/kubernetes/manifests/kube-controller-manager.manifest
     defaultconf: /etc/kubernetes/manifests/kube-controller-manager.yaml
 
   etcd:
-    confs:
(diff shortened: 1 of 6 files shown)


cfg/1.8/config.yaml
```diff
@@ -9,36 +9,13 @@
 
 master:
   apiserver:
-    confs:
-      - /etc/kubernetes/manifests/kube-apiserver.yaml
-      - /etc/kubernetes/manifests/kube-apiserver.manifest
```
lizrice: I think we still need the possibility of .manifest extension (as an alternative to .yaml) for kops and kubespray installations (see https://github.com/aquasecurity/kube-bench/commit/c44e0db97b5676c980de3dd87ba52b1a77eed478)

lizrice: (same for scheduler, controller manager & etcd below in this file)

cfg/1.8/config.yaml
```diff
@@ -9,36 +9,13 @@
...
-      - /etc/kubernetes/manifests/kube-controller-manager.manifest
     defaultconf: /etc/kubernetes/manifests/kube-controller-manager.yaml
 
   etcd:
-    confs:
-      - /etc/kubernetes/manifests/etcd.yaml
-      - /etc/kubernetes/manifests/etcd.manifest
     defaultconf: /etc/kubernetes/manifests/etcd.yaml
-
-node:
```
lizrice: Just checking - do we not need these node settings in the 1.8 version of config.yaml because we pick it up from the defaults in cfg/config.yaml?

ttousai: Yes we pick the default config from `cfg/config.yaml`

cfg/1.8/config.yaml
ttousai: I don't see anything wrong with keeping them. My idea was to clean up our configs a bit and there is no actual problem keeping them.

PR #256, opened 2019-04-09
Rationalize and document config

ttousai:
This PR documents kube-bench configuration and benchmark specification.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,4 +1,4 @@
-# Overview
+# Test and config files
 
 `kube-bench` runs checks specified in `controls` files that are a YAML 
 representation of the CIS Kubernetes Benchmark checks. There is a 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -0,0 +1,111 @@
+## Overview
+`kube-bench` runs YAML specifications of the CIS Kubernetes Benchmarks against a kubernetes cluster.
+
+## Architecture
+benchmark spec  --> kube-bench <-- configuration files
```
lizrice: I'd actually say the benchmark spec -> config files -> kube-bench -> results (rather than a direct connection between benchmark spec and kube-bench) - wdyt?

ttousai: > Is the idea here to have a separate docs file, or is this supposed to be a WIP on the main README? If the latter, let's just do PRs against the README so it's easy to read the diffs.
> 
> Having said that, I quite like the idea of separating the docs for config file syntax into a separate file that's linked from the main README.

The idea is to have a separate document file(s) linked to from README.md. This way we can keep the README short and sweet and point users to details in the `docs/` directory.

WDYT?

lizrice: That sounds good to me

lizrice:
Oh wait, this is still WIP right? So I shouldn't merge into master

ttousai:
@lizrice yes it is still WIP. I have a few more chaadditions, changes and then read over.

PR #260, opened 2019-04-11
Json & YAML config, continued

lizrice:
Building on the work that @Pluies did in #239


christopherhein:
even though the file is named `cfg/1.11-json/config.yaml` will this also work for `1.12`?

lizrice:
@christopherhein  Yes, should do

christopherhein:
Awesome, thanks @lizrice

PR #261, opened 2019-04-16
update files

yoavrotems:
Update OpenShift yaml files, to consist with RHEL guidelines

### DIFF
--- a/cfg/ocp-3.10/master.yaml
+++ b/cfg/ocp-3.10/master.yaml
@@ -1157,7 +1157,7 @@ groups:
     tests:
       bin_op: and
       test_items:
-      -  : "Binary file /proc/1/environ matches"
+      - flag: "Binary file /proc/1/environ matches"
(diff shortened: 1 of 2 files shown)


cfg/ocp-3.10/master.yaml
```diff
@@ -1,1500 +1,1454 @@
...
+      under etcdClientInfo like below.
+      
+        etcdClientInfo:
+          ca: master.etcd-ca.crt
+          certFile: master.etcd-client.crt
+          keyFile: master.etcd-client.key
+    scored: true
+
+  - id: 1.27
+    text: "Verify that the ServiceAccount admis
```
lizrice: Invalid YAML here

lizrice: Hmm, the reference is way wrong but anyway there was a missing "flag" which I have corrected.

Issue #262, opened 2019-04-17
Encryptor provider config for 1.13 use old flag


pipo02mix:
I have seen in the 1.13 master config there rule for the encryptor provide config still using `experimental` flag

```yaml
id: 1.1.34
    text: "Ensure that the --experimental-encryption-provider-config argument is
    set as appropriate (Scored)"
    audit: "ps -ef | grep $apiserverbin | grep -v grep"
    tests:
      test_items:
      - flag: "--experimental-encryption-provider-config"
        set: true
``` 

In [Kubernetes docs](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/#configuration-and-determining-whether-encryption-at-rest-is-already-enabled) it says from `1.13` the flag will be `--encryption-provider-config`

I am providing a PR for this

ttousai:
@pipo02mix thank you for catching this and the PR. 

`kube-bench` checks the `experimental-encryption-provider-config` flag to follow the [CIS kubernetes benchmark v1.4.0](https://www.cisecurity.org/benchmark/kubernetes/) recommendations as closely as possible.

In light of this we will raise this as a bug with CIS.
cc @lizrice

lizrice:
This is now fixed through the support of version 1.4.1 of the CIS spec

pipo02mix:
great

PR #263, opened 2019-04-17
Remove experimental from encryption flag

pipo02mix:
It closes the issue https://github.com/aquasecurity/kube-bench/issues/262


ttousai:
Thank you for the PR @pipo02mix though we will not be accepting it. Please see my comment on #262.

Issue #225, opened 2019-02-24
helm-chart for kube-bench


davidkarlsen:
It would be nice to have a helm-chart for running kube-bench regularly through CronJobs in a cluster - similar to https://github.com/helm/charts/tree/master/stable/kube-hunter.

lizrice:
It would. The main difference is that kube-hunter only needs to run in one location, whereas kube-bench needs to be scheduled onto every node where the tests are going to be run. We typically do that with a DaemonSet, but then that only runs once. What we need is some sort of CronJob / DaemonSet combo which would run on all nodes, on a regular basis...

davidkarlsen:
Linking some issues https://github.com/kubernetes/kubernetes/issues/36601 https://etoews.github.io/blog/2017/05/13/run-once-daemonset-on-kubernetes/

lizrice:
With https://github.com/kubernetes/kubernetes/issues/36601 being closed, I guess we are unlikely to see native support for running jobs on each node regularly. So I guess the way we can work around this is with a DaemonSet that runs a containerized regular job to run kube-bench. 

This will be easier with #72 as we wouldn't have to worry about what kind of node the job is running on.

lizrice:
Thanks to #72 we can run kube-bench on any node and it will automatically detect whether to run the master or node tests. 

We could extend the [entrypoint script](https://github.com/aquasecurity/kube-bench/blob/master/entrypoint.sh) to add a new command (say, `repeat`) that calls kube-bench every day. 

And then a DaemonSet could easily install this on every node.

lizrice:
After much debate we've reached the conclusion that this isn't something we want to support within the kube-bench project, without the underlying support of a native Kubernetes DaemonSet / CronJob combination. It's possible to do without that support, but it doesn't feel like a clean solution and we are concerned about maintaining it going forward.

subodhp:
Hi @lizrice - How do you suggest one should run the kube-bench in periodic manner in current state in k8s cluster?

Thanks,
Subodh

mirtov-alexey:
@lizrice 
Hello!
Do I understand correctly that this problem can be partially solved through the starboard?
will it launch kube-bench when a new node joins a node?
but it will not run repeated checks again on the already connected node in case someone just logs in and changes the config file? https://blog.aquasec.com/automate-kubernetes-compliance
does starboard not have this feature?

What are the other alternatives to running kube-bench on a periodic basis?

[...]

PR #272, opened 2019-04-29
[Issue 225 (enhancement)] Adding repeat option

jschalz:
Regarding [Issue 225 (enhancement): "helm-chart for kube-bench"](https://github.com/aquasecurity/kube-bench/issues/225).

`entrypoint.sh` now includes a "repeat" option for users running kube-bench in a kubernetes cluster. This will schedule a `kubectl apply -f job.yaml` command to run every 24 hours, and will run kube-bench in that deployment. Any notifications from kubectl about the job is written to a file created called `repeat-logs.txt`.

To run:
`./entrypoint.sh repeat`

**Potential issues/improvements**:

- Currently no CLI option to stop the `repeat` (but given the infrequency of the job, a `kill` command is useful, but not elegant)
- Logging needs revamping to be more verbose/helpful (rather than just job output)
- Currently only available for users running kube-bench via a kube cluster (needs a more generic solution)

**Developed with:**

- Machine: OSX
- Kubernetes version: 1.14.0
- Minikube version: 1.0.0

Thank you!

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -14,6 +14,10 @@ if [ "$1" == "install" ]; then
     echo "  run:     docker run --rm --pid=host aquasec/kube-bench [command]"
     exit
   fi
+elif [ "$1" == "repeat" ]; then
+  echo "Now scheduling kube-bench to run every 24 hours"
+  touch repeat-logs.txt
+  ./repeat-loop.sh > repeat-logs.txt &
 else
   exec kube-bench "$@"
 fi
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -16,8 +16,7 @@ if [ "$1" == "install" ]; then
   fi
 elif [ "$1" == "repeat" ]; then
   echo "Now scheduling kube-bench to run every 24 hours"
-  touch repeat-logs.txt
-  ./repeat-loop.sh > repeat-logs.txt &
+  ./repeat-loop.sh &
 else
   exec kube-bench "$@"
 fi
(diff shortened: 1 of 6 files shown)


repeat-loop.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh -e
+while true
+do
+  $(which kubectl) apply -f job.yaml
```
lizrice: I haven't tried it yet but I suspect that repeating `kubectl apply` won't actually re-run the job, because the job already exists with the same name.

jschalz:
Modified the previous `kubectl` command in `repeat-loop.sh` to now be `exec kube-bench "$@"`, per @lizrice's suggestion. This should make the solution significantly more generic!

entrypoint.sh
```diff
@@ -14,6 +14,9 @@ if [ "$1" == "install" ]; then
     echo "  run:     docker run --rm --pid=host aquasec/kube-bench [command]"
     exit
   fi
+elif [ "$1" == "repeat" ]; then
+  echo "Now scheduling kube-bench to run every 24 hours"
+  ./repeat-loop.sh &
```
lizrice: Running this in the background means that this entrypoint.sh script can exit...which then terminates the container

repeat-loop.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh -e
+while true
+do
+  exec kube-bench
```
lizrice: Need to pass the parameters through (for example if someone specifies `--version`)

lizrice: To try this out I had to add the repeat-loop.sh to the Dockerfile

PR #273, opened 2019-04-29
Add flags to further filter CIS checks to run

### LINKED ISSUE #172: Add flags to further filter CIS checks to run
I think the following flags will be useful for further filtering which checks to run:
```
--scored:      to run only CIS scored checks.
--unscored:  to run only unscored CIS checks.
--level <n>:  to run only checks for a certain CIS level.
```

danielpacak:
Changes:

1. Partially resolves #172 by adding support for `--scored` and `--unscored` flags
2. Some refactoring to facilitate unit testing of new/existing functionality.

Tested with unit tests and as well as minikube v1.0.0

### DIFF
--- a/check/check.go
+++ b/check/check.go
@@ -36,11 +36,11 @@ const (
 	// PASS check passed.
 	PASS State = "PASS"
 	// FAIL check failed.
-	FAIL = "FAIL"
+	FAIL State = "FAIL"
 	// WARN could not carry out check.
-	WARN = "WARN"
+	WARN State = "WARN"
 	// INFO informational message
-	INFO = "INFO"
+	INFO State = "INFO"
 
 	// MASTER a master node
 	MASTER NodeType = "master"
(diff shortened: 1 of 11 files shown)


danielpacak:
@lizrice @jerbia can you have a look plz?

cmd/root.go
```diff
@@ -79,16 +85,18 @@ func init() {
 	RootCmd.PersistentFlags().BoolVar(&noRemediations, "noremediations", false, "Disable printing of remediations section")
 	RootCmd.PersistentFlags().BoolVar(&jsonFmt, "json", false, "Prints the results as JSON")
 	RootCmd.PersistentFlags().BoolVar(&pgSQL, "pgsql", false, "Save the results to PostgreSQL")
+	RootCmd.PersistentFlags().BoolVar(&filterOpts.Scored, "scored", false, "Run only scored CIS checks")
+	RootCmd.PersistentFlags().BoolVar(&filterOpts.Unscored, "unscored", false, "Run only unscored CIS checks")
```
lizrice: I'm not convinced by the naming / semantics of these options. If you were to set both `scored` and `unscored` flags to true, then that would appear to be saying "run only scored checks AND run only unscored checks" which is paradoxical! I haven't checked what the logic in the code actually does but perhaps a better way of phrasing this would be: 

- `scored` defaults to true, and means `run the scored checks`
- `unscored` defaults to true, and means `run the unscored checks`

If you specify neither of the flags, all the checks get run (backwards compatible)
If you specify either of the flags to be true, that makes no difference as it matches the default
if you specify one of the flags to be false, that set of checks is not run (so if you specify both the flags to be false, none of the checks are run). 

Wdyt?

check/check.go
```diff
@@ -74,20 +74,37 @@ type Check struct {
 	Scored      bool   `json:"scored"`
 }
 
+// Runner wraps the basic Run method.
```
lizrice: this makes me happy 💯

cmd/root.go
danielpacak: Thanks for pointing this out @lizrice . The logic in the code is indeed paradoxical. I really like the semantics you proposed. Let me implement it. Should be quick.

danielpacak: Should work as specified now

lizrice:
Thanks @danielpacak!

PR #281, opened 2019-05-05
The check's actual result property is now set to be the audit command…

### LINKED ISSUE #280: Actual value is empty
The "actual_value" field in the program's output is empty

yoavAqua:
…'s output

fix #280

### DIFF
--- a/check/test.go
+++ b/check/test.go
@@ -135,7 +135,6 @@ func (t *testItem) execute(s string) *testOutput {
 				}
 			}
 
-			result.actualResult = strings.ToLower(flagVal)
 			switch t.Compare.Op {
 			case "eq":
 				value := strings.ToLower(flagVal)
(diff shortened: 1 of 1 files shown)


check/test.go
```diff
@@ -230,7 +229,7 @@ func (ts *tests) execute(s string) *testOutput {
 	}
 
 	finalOutput.testResult = result
-	finalOutput.actualResult = res[0].actualResult
+	finalOutput.actualResult = s
```
lizrice: This looks strange to me - isn't `s` the audit command to be executed?

yoavAqua: it's the output of the audit command

lizrice: Ah, OK... i can't help thinking that the name `execute()` is pretty bad for these functions - more like `checkResults()`. 

On the occasions where we call `failTestItem()` the string is currently being set to an error message explaining what's wrong (e.g. bad YAML). I think with this change we are going to lose those messages as they're going to get overwritten by the original string. Wdyt?

yoavAqua: True, fixing it now

yoavAqua:
@lizrice Can you please review? I fixed the issue with overriding the failTestItem() output by first checking that the actual result is empty.

PR #282, opened 2019-05-05
Printing the actual test result of failed tests - when a flag is raised

### LINKED ISSUE #110: Show test result as well as desired result
We show that a test has failed and we show what the remediation is, but it would be useful to also show the actual test result when it fails.

yoavAqua:
fix #110


yoavAqua:
@lizrice Can you please merge and close this pr?

PR #284, opened 2019-05-06
Genereate expected result automatically for each test

yoavAqua:
The expected result of each test, is generated from the test arguments from the yaml file.
Every expected result is a combination of expressions, separated by the binary operand (or/and).
Every expression is combined from the test item attributes:
flag, set and if there is - compare operand and compare value.

Example:
tests:
bin_op: or
test_items:
- flag: "--cgroup-parent"
set: false
- flag: "--cgroup-parent"
compare:
op: nothave
value: "/docker"
set: true
Expected result will be shown as: '--cgroup-parent' Is not present OR '--cgroup-parent' Not have '/docker'

Another example:
tests:
test_items:
- flag: "userns"
compare:
op: has
value: "userns"
set: true
Expected result will be shown as: 'userns' Has 'userns'

### DIFF
--- a/check/test.go
+++ b/check/test.go
@@ -219,8 +219,11 @@ func (ts *tests) execute(s string) *testOutput {
 		return finalOutput
 	}
 
+	expectedResultArr := make([]string, len(res))
+
 	for i, t := range ts.TestItems {
 		res[i] = *(t.execute(s))
+		expectedResultArr[i] = res[i].ExpectedResult
 	}
 
 	var result bool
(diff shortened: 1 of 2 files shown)


check/test.go
```diff
@@ -135,9 +136,11 @@ func (t *testItem) execute(s string) *testOutput {
 				}
 			}
 
+			expectedResultPattern := ""
 			result.actualResult = strings.ToLower(flagVal)
 			switch t.Compare.Op {
 			case "eq":
+				expectedResultPattern = "'%s' Is equal to '%s'"
```
lizrice: I think all these should be lower case `is` rather than `Is` (same for all the other expected result strings)

check/test.go
```diff
@@ -220,13 +233,19 @@ func (ts *tests) execute(s string) *testOutput {
 	case and, "":
 		result = true
 		for i := range res {
+			finalOutput.ExpectedResult += fmt.Sprintf("%s AND ", res[i].ExpectedResult)
```
lizrice: Use [strings.Join()](https://golang.org/pkg/strings/#Join) for these, it's neater and means you don't have to manually delete the last iteration

yoavAqua: You are right, but this for loop will still happen so it makes it not efficient since we're iterating on the array 2 times

yoavAqua: I can go with whatever you prefer, what do you say?

lizrice: strings.Join() is (a) idiomatic (b) [almost certainly more efficient than roll-your-own string concatenations with fmt.Printf](https://syslog.ravelin.com/bytes-buffer-i-thought-you-were-my-friend-4148fd001229).

yoavAqua: Wow amazing! you really blow my mind right now :)
So the only problem I see now is that I need to join not the res array but the ExpectedResult property of each item in that array.
What I can do now is either keep this code or maybe append the ExpectedResult of each item to a new array, and then call the strings.Join() method.
What do you think?

lizrice: Good point - I'd probably do it something like [this](https://play.golang.org/p/8uh_FW1MK53)

yoavAqua: Done

yoavAqua:
@lizrice please merge

Issue #235, opened 2019-03-06
Rationalize and document config files


lizrice:
We should document how the config.yaml files work and how cfg/config.yaml and cfg/<version>/config.yaml are merged. I think that any settings in the version-specific config file override what's in the main config file, right? 

Further, I think we can make the config files simpler going forward, and make it work for more installers. As an example, let's consider the settings we currently have for the API Server config file. 

* In the main config.yaml we have some alternatives `/etc/kubernetes/apiserver.conf` and `/etc/kubernetes/apiserver` but I'm pretty sure those haven't been used for a long time (at least not since 1.8). 
* In 1.8/config.yaml we introduce the alternatives `/etc/kubernetes/manifest/kube-apiserver.yaml` and `/etc/kubernetes/manifest/kube-apiserver.manifest` where the .manifest extension is used by kops and kubespray, and the .yaml extension is used by Kubeadm. 
* In 1.11/config.yaml we only have `/etc/kubernetes/manifest/kube-apiserver.yaml`, so I don't see how this would work for a Kops or kubespray installation? 
* We just copied the 1.11 version for 1.13

If I'm not misunderstanding how this works, I think we could make this simpler by having the following 
* In the main config.yaml, use what we currently have in 1.8/config.yaml i.e. `/etc/kubernetes/manifest/kube-apiserver.yaml` and `/etc/kubernetes/manifest/kube-apiserver.manifest` with a default of `/etc/kubernetes/manifest/kube-apiserver.yaml`
* We would then have no need for apiserver conf file settings in 1.8/config.yaml, 1.11/config.yaml or 1.13/config.yaml.  They could be deleted as the setting would be picked up from the main config.yaml file
* We could copy the old settings (currently in the main config.yaml) into 1.7 and 1.6/config,yaml so that they continue to use those old locations. 

The same approach could apply to several other file settings. Ideally we'd end up in a situation where we don't need version-specific config at all, because the config can all be picked up from the main file.

daniellohausen:
A remark :

> In 1.11/config.yaml we only have /etc/kubernetes/manifest/kube-apiserver.yaml, so I don't see how this would work for a Kops or kubespray installation

I does not work, that's what I stumbled upon today. I can create a bug or pull request to add the kops file destination to the 1.11 config.

lizrice:
Sorry for slow response here @daniellohausen - if you have a PR for this that would be terrific

lizrice:
Thanks @daniellohausen. 

[...]

PR #296, opened 2019-05-30
Document version-specific config files

lizrice:
Values in the version-specific files override the main file.

I think this is the last outstanding item to-do for #235


lizrice:
I think this is important enough we should cover it in the main README

PR #297, opened 2019-05-30
Update openshift executables

lizrice:
As recommended by @bgoareguer in https://github.com/aquasecurity/kube-bench/issues/242#issuecomment-494722911


yoavrotems:
LGTM ;)

PR #299, opened 2019-06-04
remove removed --repair-malformed-updates check

nikopen:
this has been removed from 1.13 onwards (was already deprecated)

(any hint on updating `id` numbers?)

### DIFF
--- a/cfg/1.13/master.yaml
+++ b/cfg/1.13/master.yaml
@@ -133,22 +133,6 @@ groups:
       --profiling=false
     scored: true
 
-  - id: 1.1.9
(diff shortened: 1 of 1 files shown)


ttousai:
@nikopen thank you for the PR.

Unfortunately we can't accept this PR because the CIS Kubernetes Benchmark v1.4.0 requires this check, although it may be deprecated.

We aim to match the Benchmark as closely as possible and so we will retain this check.

lizrice:
@nikopen I have checked, and the next version of the CIS benchmark will not include this test. 

In the mean time you can modify your own local copy of the test either to delete the test, or to mark it as `type: skip` as described [here](https://github.com/aquasecurity/kube-bench#omitting-checks)

Issue #303, opened 2019-06-05
Remove cipher suites TLS_RSA_WITH_AES_256_GCM_SHA384 and TLS_RSA_WITH_AES_128_GCM_SHA256


zhedahht:
TLS_RSA_WITH_AES_256_GCM_SHA384 and TLS_RSA_WITH_AES_128_GCM_SHA256 have been marked as weak cipher suites by SSL Labs, and we should not include them in tls-cipher-suites.

simar7:
Thanks @zhedahht I've added a PR to tackle this https://github.com/aquasecurity/kube-bench/pull/308

lizrice:
Sorry to be slow to respond on this. Because we follow the CIS Kubernetes Benchmark we need these cipher suites to be removed from the benchmark before we remove them from kube-bench (though you can of course modify the test yaml that you run with in your own environment in the mean time). I am raising a ticket on the CIS workbench about it.

lizrice:
Ticket raise here: https://workbench.cisecurity.org/benchmarks/1370/tickets/8550

Issue #302, opened 2019-06-05
option "tls-cipher- suites" should be "tls-cipher-suites"


zhedahht:
In https://github.com/aquasecurity/kube-bench/blob/master/cfg/1.13/master.yaml#L505, the option "tls-cipher- suites" should be "tls-cipher-suites", removing the space between the second '-' and "suites".

In the same line, we should also remove extra spaces from the following word:
TLS_ECDHE_RSA_WITH_AES_128_GCM _SHA256
TLS_ECDHE_RSA_WITH_AES_256_GCM _SHA384
TLS_ECDHE_ECDSA_WITH_AES_256_GCM _SHA384

yoavrotems:
Due to https://github.com/aquasecurity/kube-bench/pull/308 i think we can close it here and move the discussion to there

PR #308, opened 2019-06-10
remove weak cipher suites

simar7:
This commit removes recently classified weak cipher suites:
`TLS_RSA_WITH_AES_256_GCM_SHA384`
`TLS_RSA_WITH_AES_128_GCM_SHA256`

Handles the following: 
Removing weak ciphers: https://github.com/aquasecurity/kube-bench/issues/303
Erroneous spacing in `tls-cipher-suites`: https://github.com/aquasecurity/kube-bench/issues/302 

Ref: https://github.com/ssllabs/research/wiki/SSL-and-TLS-Deployment-Best-Practices#23-use-secure-cipher-suites

Signed-off-by: Simarpreet Singh <simar@linux.com>


yoavrotems:
About the extra spaces, you are correct.
About removing the other parts from the remediation 
> TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256

I'm not so sure about it, because we are sticking to CIS books, and that is the recommendation for those, @lizrice wdyt?

simar7:
@lizrice that makes sense! I'll cut a new PR here https://github.com/aquasecurity/kube-bench/pull/321 and close this one in the interest of that.

PR #326, opened 2019-06-26
Detect kube-controller in CMD

mrueg:
If kube-controller-manager is getting detected by older versions of
procps, it will only be detected if we're looking for kube-controller
(15 chars)

NOTE: "The command name is not the same as the command line. Previous versions of
       procps and the kernel truncated this command name to 15
       characters. This limitation is no longer present in both. If
       you depended on matching only 15 characters, you may no longer
       get a match."


mrueg:
Yeah the other option would be to run pgrep -a $controllerbinary so it hits the full command line. But as of right now, we should be good with checking for kube-controller.

PR #330, opened 2019-06-28
fix go.mod dependency

patelpayal:
Due to go mod pointing to`	github.com/aquasecurity/kube-bench v0.0.29	` at line#6, it does not honor the current code and points to the dependency cache.


lizrice:
Agh, sorry I missed this in previous change

PR #336, opened 2019-07-02
Add codecov

yoavrotems:
adding the codecov feature to project


makefile
```diff
@@ -33,6 +33,7 @@ build-docker:
 
 tests:
 	go test -race -timeout 30s -cover ./cmd ./check
```
danielpacak: I think this line is redundant. In the line below you are running tests in all packages, including ./cmd and .check.

.travis.yml
```diff
@@ -13,17 +13,28 @@ before_install:
...
+  - go get -t -v ./...
 
 script:
   - GO111MODULE=on go test ./...
   - docker build --tag kube-bench .
   - docker run -v `pwd`:/host kube-bench install 
   - test -d cfg
   - test -f kube-bench
-
+  - go test -race -coverprofile=coverage.txt -covermode=atomic ./...
```
danielpacak: I'd run `make tests` instead of duplicating tests execution command with args. So you and travis can use the same command to run tests. Pretty useful for troubleshooting failing tests locally.

makefile
yoavrotems: Done

Issue #243, opened 2019-03-15
Cipher suites warning after I set the specific ciphers


horis233:
I got the warning message after I used the cipher suites provided by the remediation messages.
Command`ps -ef |  grep -v grep | grep tls-cipher-suites`

[...]

PR #357, opened 2019-07-11
issue #243: Changes condition so that score: false tests are performed

robertojrojas:
Changes condition so that score: false tests are performed.

### DIFF
--- a/check/check.go
+++ b/check/check.go
@@ -102,7 +102,7 @@ func (c *Check) run() State {
 		return c.State
 	}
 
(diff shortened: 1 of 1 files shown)


check/check.go
```diff
@@ -103,7 +103,7 @@ func (c *Check) run() State {
 	}
 
 	// If check type is manual or the check is not scored, force result to WARN
-	if c.Type == "manual" || !c.Scored {
+	if c.Type == "manual" {
```
lizrice: Should change the comment too

Issue #344, opened 2019-07-10
Kubelet Config: Add Support for Test Comparison of Array Values


robertojrojas:
The value of the property **'tlsCipherSuites'** Kubelet Configuration file is an array of strings.
The tests does not appear to be working while parsing this property and reporting that 
the property is not being set correctly.
Here is a snippet sample of the Kubelet config file:

`{
  ...
   "tlsCipherSuites": [
    "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256",
    "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"
  ],
  ...
}
`

This is the JSONPATH string in the **cfg/<version>-json/node.yaml** used to find the Kubelet config property value. Unfortunately, it doesn't appear to be working:

`
...
tests:
      test_items:
      - path: "{.tlsCipherSuites}"
        compare:
          op: eq
...
`

robertojrojas:
This issue appears to be related to issue #243

lizrice:
@robertojrojas is this done?

robertojrojas:
yes, it was fixed by #367

PR #367, opened 2019-07-15
issue #344: Adds support for array comparison. Every element in the s…

robertojrojas:
Adds support for array comparison. Every element in the `source` array must exist in the `target` array. An empty `source` array is reported as invalid.

### DIFF
--- a/check/test.go
+++ b/check/test.go
@@ -37,8 +37,9 @@ import (
 type binOp string
 
 const (
-	and binOp = "and"
-	or        = "or"
+	and                   binOp = "and"
+	or                          = "or"
+	defaultArraySeperator       = ","
 )
 
 type testItem struct {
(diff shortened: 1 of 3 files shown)


check/test.go
```diff
@@ -37,8 +37,9 @@ import (
 type binOp string
 
 const (
-	and binOp = "and"
-	or        = "or"
+	and                   binOp = "and"
+	or                          = "or"
+	defaultArraySeperator       = ","
```
lizrice: Nit: spelling `defaultArraySep`**a**`rator`

check/test.go
```diff
@@ -220,6 +228,33 @@ func executeJSONPath(path string, jsonInterface interface{}) (string, error) {
...
+	}
+	for _, sv := range s {
+		found := false
+		for _, tv := range t {
+			if sv == tv {
+				found = true
+				break
+			}
+		}
+		if found == false {
```
lizrice: I think I'd say `if !found` rather than `if found == false`

Issue #372, opened 2019-07-23
Deduplicate documentation


ttousai:
We have some info in the main README that is better documented in docs/README.md. Update the main README to remove the duplicate info and refer to the docs/README.md instead?

PR #373, opened 2019-07-23
Remove duplicate documentation.

ttousai:
We have some info in the main README that is better documented in docs/README.md. Update the main README to remove the duplicate info and refer to the docs/README.md instead?

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,8 @@ There are three output states
 - [WARN] means this test needs further attention, for example it is a test that needs to be run manually 
 - [INFO] is informational output that needs no further action.
 
+## Test config YAML representation
+
 ### Omitting checks
 
 If you decide that a recommendation is not appropriate for your environment, you can choose to omit it by editing the test YAML file to give it the check type `skip` as in this example: 
--- a/README.md
+++ b/README.md
@@ -142,12 +142,7 @@ Kubernetes config and binary file locations and names can vary from installation
 
 Any settings in the version-specific config file `cfg/<version>/config.yaml` take precedence over settings in the main `cfg/config.yaml` file.
 
-For each type of node (*master*, *node* or *federated*) there is a list of components, and for each component there is a set of binaries (*bins*) and config files (*confs*) that kube-bench will look for (in the order they are listed). If your installation uses a different binary name or config file location for a Kubernetes component, you can add it to `cfg/config.yaml`.
-
-* **bins** - If there is a *bins* list for a component, at least one of these binaries must be running. The tests will consider the parameters for the first binary in the list found to be running.
-* **podspecs** - From version 1.2.0 of the benchmark (tests for Kubernetes 1.8), the remediation instructions were updated to assume that the configuration for several kubernetes components is defined in a pod YAML file, and podspec settings define where to look for that configuration.
-* **confs** - If one of the listed config files is found, this will be considered for the test. Tests can continue even if no config file is found. If no file is found at any of the listed locations, and a *defaultconf* location is given for the component, the test will give remediation advice using the *defaultconf* location.
-* **unitfiles** - From version 1.2.0 of the benchmark  (tests for Kubernetes 1.8), the remediation instructions were updated to assume that kubelet configuration is defined in a service file, and this setting defines where to look for that configuration.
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -202,45 +165,6 @@ If you decide that a recommendation is not appropriate for your environment, you
...
-- `noteq`: tests if the flag value is unequal to the compared value.
-- `gt`: tests if the flag value is greater than the compared value.
-- `gte`: tests if the flag value is greater than or equal to the compared value.
-- `lt`: tests if the flag value is less than the compared value.
-- `lte`: tests if the flag value is less than or equal to the compared value.
-- `has`: tests if the flag value contains the compared value.
-- `nothave`: tests if the flag value does not contain the compared value.
-- `regex`: tests if the flag value matches the compared value regular expression.
-
 When defining regular expressions in YAML it is generally easier to wrap them in single quotes, for example `'^[abc]$'`, to avoid issues with string escaping.
```
lizrice: The sentence about defining regex's should go where the `regex` operation is described, it doesn't make sense just left here. Also, the definition of `regex` is currently missing from the list in the separate [docs directory](https://github.com/aquasecurity/kube-bench/blob/master/docs/README.md).

README.md
```diff
@@ -156,38 +151,6 @@ There are three output states
 - [WARN] means this test needs further attention, for example it is a test that needs to be run manually 
 - [INFO] is informational output that needs no further action.
 
-## Test config YAML representation
```
lizrice: I think we should still have at least this heading and the initial sentence in the main readme, so people can easily see that tests are configured in YAML. And there should be a link to take them to the details. We probably need the example as well, because otherwise the next section on Omitting checks doesn't make sense.

README.md
```diff
@@ -156,38 +151,6 @@ There are three output states
...
-        set: true
-      - flag: "--some-other-flag"
-        set: false
-      remediation: "Edit the /etc/kubernetes/config file on the master node and set the KUBE_ALLOW_PRIV parameter to '--allow-privileged=false'"
-      scored: true
-```
-
-Recommendations (called `checks` in this document) can run on Kubernetes Master, Node or Federated API Servers.
-Checks are organized into `groups` which share similar controls (things to check for) and are grouped together in the section of the CIS Kubernetes document.
-These groups are further organized under `controls` which can be of the type `master`, `node` or `federated apiserver` to reflect the various Kubernetes node types.
```
lizrice: The type should be `federated` not `federated apiserver` here

ttousai:
> While you are correcting this, please could you also look at the text in [docs/README.md](https://github.com/aquasecurity/kube-bench/blob/master/docs/README.md) that describes how bins, confs etc are substituted? I am not sure that it's correct about a few things:
>
> * Aren't `defaultbin`, `defaultconf` etc the string that we substitute if we can't find one of `bins` running / one of the named files in `confs`? I think that's the behaviour, but it's not what the text says.

Your statement is correct, and it says the same in docs/README. The description may be wordy but I think it conveys the same fact in great detail. For example the `bins` configuration section describes 3 things:
- the `bins` options and what it does
- the `defaultbin` option, when it is considered or ignored
- what happens if the binary specified by either `bins` or `defaultbin` is not running on the host.

One change I have to make is to remove mention of the fact that kube-bench terminates if `bins` is not running, that should only be mentioned after `defaultbin` is checked.

---

> * And if that's correct, "kube-bench terminates if [this file does not exist / this binary is not running]" is not true in all the places it is said in the docs text.

I see some problem in the text here. I will change it to reflect this. I have confirmed in code that kube-bench only terminates if binaries specified by `bins` or `defaultbins` is not running, but runs for missing configs, service files etc.

---
 
> The text at the beginning of the [docs/README.md](https://github.com/aquasecurity/kube-bench/blob/master/docs/README.md) file talks about auto-detecting or specifying the version - this should really be in the main README as it's something you need to know when running kube-bench.

I will move it there.

README.md
ttousai: Done

README.md
ttousai: Done.

README.md
ttousai: I deleted this in the main README but did not add any mention of federation in docs/README.

lizrice:
Fixed #372

Issue #374, opened 2019-07-23
Check 1.1.10 fails when flag is not set


mwwolters:
The AlwaysAdmit admission controller is not enabled by [default](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#which-plugins-are-enabled-by-default), but when this check runs and the flag is not set the test fails.

yoavrotems:
I believe @robertojrojas is fixing this issue right?

mwwolters:
I think #375 should fix this, @robertojrojas if you were working on this sorry to step on your toes.

robertojrojas:
@mwwolters it's  all good. All contributions are really appreciated 💯

PR #375, opened 2019-07-23
Updated check to pass if flag isn't set

mwwolters:
#374 : The default for --enable-admission-plugins does not include AlwaysAdmit, changed this check to also pass if the flag isn't set.

### DIFF
--- a/cfg/1.11/master.yaml
+++ b/cfg/1.11/master.yaml
@@ -153,12 +153,15 @@ groups:
     text: "Ensure that the admission control plugin AlwaysAdmit is not set (Scored)"
     audit: "ps -ef | grep $apiserverbin | grep -v grep"
     tests:
+      bin_op: or
       test_items:
       - flag: "--enable-admission-plugins"
         compare:
           op: nothave
           value: AlwaysAdmit
         set: true
+      - flag: "--enable-admission-plugins"
+        set: false
     remediation: |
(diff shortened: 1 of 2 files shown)


mwwolters:
The flag [comment](https://github.com/kubernetes/kubernetes/blob/016e9d5c06089774c6286fd825302cbae661a446/staging/src/k8s.io/apiserver/pkg/server/options/admission.go#L97) says any set are in addition to default enabled ones. My reading of the [ApplyTo](https://github.com/kubernetes/kubernetes/blob/016e9d5c06089774c6286fd825302cbae661a446/staging/src/k8s.io/apiserver/pkg/server/options/admission.go#L116) function (which adds the admissions chain to the server config) is that it calls [enabledPluginNames](https://github.com/kubernetes/kubernetes/blob/016e9d5c06089774c6286fd825302cbae661a446/staging/src/k8s.io/apiserver/pkg/server/options/admission.go#L204) which I think takes the flag value, generates a list of disabled plugins based on the default off and non-flag plugins and creates a list of the enabled ones by selecting those not in the list of disabled ones. Does that look right?

lizrice:
Thank you @mwwolters for the extra links, I have re-read and now I agree with you!

mwwolters:
@lizrice no problem, thanks for taking a look!

PR #376, opened 2019-07-25
fixes-according-kube-cis1.4.1

yoavrotems:
test 1.1.12 got DEPRECATED.
test 1.1.34 changed to --encryption-provider-config and need to be set manually

### DIFF
--- a/cfg/1.13/master.yaml
+++ b/cfg/1.13/master.yaml
@@ -183,9 +183,8 @@ groups:
     scored: true
 
   - id: 1.1.12
-    text: "Ensure that the admission control plugin DenyEscalatingExec is set (Scored)"
+    text: "[DEPRECATED] Ensure that the admission control plugin DenyEscalatingExec is set (Not Scored)"
     audit: "ps -ef | grep $apiserverbin | grep -v grep"
-    # [DEPRECATED]
     type: skip
     tests:
       test_items:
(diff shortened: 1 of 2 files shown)


cfg/1.13/master.yaml
```diff
@@ -185,6 +185,8 @@ groups:
   - id: 1.1.12
     text: "Ensure that the admission control plugin DenyEscalatingExec is set (Scored)"
```
lizrice: I think it’s worth also editing the text field to add “[DEPRECATED]” like the benchmark document

cfg/1.13/master.yaml
```diff
@@ -556,19 +558,22 @@ groups:
...
+    text: "Ensure that the --encryption-provider-config argument is set as appropriate (Scored)"
     audit: "ps -ef | grep $apiserverbin | grep -v grep"
+    type: "manual"
     tests:
       test_items:
-      - flag: "--experimental-encryption-provider-config"
+      - flag: "--encryption-provider-config"
+        compare:
+          op: has
+          value: "</path/to/EncryptionConfig/File>"
```
lizrice: “</path/to/EncryptionConfig/File>” isn’t supposed to be a real location, it could be wherever the user has put the config file, and there is no default for this. That’s why we were just testing that the field is set, but not checking what value it’s set to. (I suppose we could be checking that it’s not set to an empty string.)

cfg/1.13/node.yaml
```diff
@@ -222,6 +222,8 @@ groups:
   - id: 2.1.11
     text: "Ensure that the --cadvisor-port argument is set to 0 (Scored)"
```
lizrice: As above, it would be good to add [DEPRECATED] into the text description to match the benchmark

cfg/1.13/master.yaml
yoavrotems: Done

cfg/1.13/master.yaml
yoavrotems: OK Changed it :)

PR #377, opened 2019-07-29
Fix version command failing because of missing config file it does not need.

### LINKED ISSUE #334: kube-bench version cmd should not require config dir to be set
This happens if the config is not available on the right path, but kube-bench version should work independently of that:

```
/usr/bin/kube-bench version                                                                      :(
[FAIL] Failed to read config file: Config File "config" Not Found in "[/$PATH/cfg]"
```
with a path to config set:
```
/usr/bin/kube-bench version -D /etc/kube-bench/cfg                                                 :(
0.0.30
```

ttousai:
Fixes #334


cmd/common.go
```diff
@@ -65,6 +65,12 @@ func NewRunFilter(opts FilterOpts) (check.Predicate, error) {
 func runChecks(nodetype check.NodeType) {
 	var summary check.Summary
 
+	// Verifiy config file was loaded into Viper during Cobra sub-command initialization.
```
mrueg: Verify*

PR #383, opened 2019-08-15
Update config.yaml to add Azure AKS file locations for kubelet

PatrickSplice:


cfg/config.yaml
```diff
@@ -78,15 +78,26 @@ node:
     - kubernetes
 
   kubernetes:
-    defaultconf: /etc/kubernetes/config    
+    confs:
+      - "/var/lib/kubelet/kubeconfig"
```
lizrice: kubernetesconf is only referred to in the test config file for 1.7 so I don't think this change can have any effect, so I'd prefer not to make it. (in fact I think we should be removing this section altogether and I've raised #384 for it).  This does look like a kubelet file anyway, and I think the effect you're looking for has likely been achieved by adding it into the kubelet section below

cfg/config.yaml
```diff
@@ -78,15 +78,26 @@ node:
...
       - "/etc/kubernetes/pki/ca.crt"
+      - "/etc/kubernetes/certs/ca.crt"
+    svc:
+      - "/etc/systemd/system/kubelet.service.d/10-kubeadm.conf"
+      - "/etc/systemd/system/kubelet.service"
     bins:
       - "hyperkube kubelet"
       - "kubelet"
+    kubeconfig:
+      - "/var/lib/kubelet/kubeconfig"
```
lizrice: Since the file locations are looked at in order, I would prefer the newly-added location to be added at the end of the list (i.e. add `/var/lib/kubelet/kubeconfig` after `/etc/kubernetes/kubelet.conf`) as it reduces the risk of this fix breaking another platform.

cfg/config.yaml
```diff
@@ -78,15 +78,26 @@ node:
...
+      - "/etc/kubernetes/kubelet.conf"
     confs:
+      - "/etc/default/kubelet"
```
lizrice: Similarly here, please add `/etc/default/kubelet` at the end of the list of confs. This would break an Ubuntu kubeadm install as there is a `/etc/default/kubelet` file as well as `/var/lib/kubelet/config.yaml`, and it's the latter that kube-bench should be using here

PatrickSplice:
Adjustments made.

Issue #370, opened 2019-07-18
Update 1.13 specs for new CIS release 1.4.1


lizrice:
There is an update v1.4.1 of the CIS Benchmark for Kubernetes 1.13. We should update both 1.13 and 1.13-json configs to match.

lizrice:
@yoavrotems working on this under PR #376

lizrice:
@yoavrotems #376 covers the 1.13 config, could you also do one for 1.13-json please?

cjschaef:
Does kube-bench now support the v1.4.1 CIS benchmark, given #376 has merged?

Can the README be updated now if this is the case?

lizrice:
We still need to update 1.13-json, but I agree we could update the README already

PR #385, opened 2019-08-21
Updated version support

lizrice:
Strictly, we don't have the changes in 1.13-json but we do have them in 1.13. 

Relates to #370


yoavrotems:
LGTM

PR #388, opened 2019-08-22
Updated cfg/1.13-json/node.yaml to fix 2.2.7 check for kubeletcafile"

PatrickSplice:


cfg/config.yaml
```diff
@@ -1,5 +1,5 @@
 ---
```
robertojrojas: All the changes as part of this file have already been applied to master by other PRs, so it can be safely removed as part of this PR.

lizrice:
I'm not sure what order things have happened in, but I think the changes for this are now in master (possibly merged under #523). @PatrickSplice does what's in master resolve the issue for you?

Issue #158, opened 2018-08-30
kube-bench does not support snap based kubernetes installations


CalvinHartwell:
Hi all,

The kube-bench scans do not currently detect snap based installations, the paths for the systemd units and the args are in different locations. This messes up the kubelet and master scans for Canonical/Ubuntu based installations of Kubernetes. 

I will put together a proper list of bugs and raise a PR. 

Cheers,

- Calvin

ttousai:
Hello @CalvinHartwell ! Looking forward to the PR.

CalvinHartwell:
More information on snap based package structure: [https://docs.snapcraft.io/snaps/structure](https://docs.snapcraft.io/snaps/structure), I'm hoping to get this done this week or next week to fix Aqua support for CDK.

nicorikken:
@sabbour Indeed, this is a broader issue. We run kube-proxy, kubelet and kubeapiserver in containers (Docker/Rocket) on the nodes, which also messes up running kube-bench, because the binaries are not available. Despite the containerized processes being available through `ps`. We ended up adding the binaries as dummies to trick kube-bench into executing 😆

lizrice:
Ah, thanks @nicorikken, that's an interesting clue! Could you expand on what you mean by "adding the binaries as dummies"? 

Adding the file locations and binary names into the config files *should* just work if the processes are available through ps (modulo whether ps itself is available). Also, the dependency on finding either kubectl or kubelet executable can be removed by specifying a particular version config with --version.

PR #389, opened 2019-08-22
Config for snap-based clusters

kwmonroe:
This PR adds a cfg subdirectory to support clusters based on snap-installed components (first noted back in https://github.com/aquasecurity/kube-bench/issues/158).

I split this out into a new subdir because snap-based components felt sufficiently different than kubeadm components. For example, k8s snaps rely on systemd services and `$SNAP_DATA` config files instead of manifests and `/etc/kubernetes/*.conf`.

It may be possible to work out the differences such that snaps could be supported in the `1.13` cfg dir.  The vast majority of differences are just in the remediation wording, e.g. in `master.yaml`:

```diff
-      Edit the API server pod specification file $apiserverconf
+      Edit the API server snap config file $apiserverconf
```

Similarly in `node.yaml`:

```diff
-      If using command line arguments, edit the kubelet service file
-      $kubeletsvc on each worker node and
-      set the below parameters in KUBELET_CERTIFICATE_ARGS variable.
+      If using command line arguments, edit the kubelet snap config file
+      $kubeletconf on each worker node and
+      set the below parameters.
```

That said, there are a few master tests that are skipped (anything testing for `/etc/kubernetes/*.conf`) as well as a few node tests that should be considered manual (`kube-bench` would need to be run as root to find kubeconfig and certs).

Thoughts on the best approach to support snap-based clusters?

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,10 @@ go build -o kube-bench .
 
 kube-bench includes a set of test files for Red Hat's OpenShift hardening guide for OCP 3.10 and 3.11. To run this you will need to specify `--version ocp-3.10` when you run the `kube-bench` command (either directly or through YAML). This config version is valid for OCP 3.10 and 3.11. 
 
+## Running on snap installed clusters
+
+Cluster components installed with snap packages utilize systemd services and files that reside in various snap-related directories. kube-bench includes support for these types of clusters in the `cfg/1.13-snap` subdirectory. To use this configuration, specify `--version 1.13-snap` when you run the `kube-bench` command.
+
 ## Output
 
 There are three output states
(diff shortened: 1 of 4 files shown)


lizrice:
Thank you so much for you work on this @kwmonroe!

I am a bit worried about the proliferation of lots of different config files that we'll need to maintain going forward. @robertojrojas any thoughts?  

> a few node tests that should be considered manual (kube-bench would need to be run as root to find kubeconfig and certs).

This is true on other systems too as there are often files that are only accessible to root. If you can't run as root you can run in a container, which is effectively root. (Hmm, maybe we need to document that point better...)

kwmonroe:
@lizrice @robertojrojas thank you both for the comments. I'm closing this out in favor of smaller PRs and issues to differentiate snap-specific changes from the effort to make the tests more generic (#414 and #415).

Issue #396, opened 2019-08-27
2.1.14 incorrectly uses $kubeletconf instead of $kubeletsvc


lizrice:
The remediation text explicitly says the words "kubelet service file" but then goes on to use $kubeletconf. The substitution parameter for the service file is $kubeletsvc, which should be used instead. 

I haven't checked but I expect this needs to be fixed in 1.11, 1.11-json, 1.13 and 1.13-json

lizrice:
I'd rather leave the text the same as it is in the CIS benchmark as far as possible

lizrice:
According to #396 it's correct in 1.11/node.yaml and 1.11-json/node.yaml already

lizrice:
fixed by #399

PR #399, opened 2019-08-28
Fixes Issue #396 - Replaces $kubeletconf for $kubeletsvc

robertojrojas:
Fixes issue #396   
Replaces `$kubeletconf ` for `$kubeletsvc `


robertojrojas:
That's a good questions. I'm kind of on the fence on this. If we leave the *-json.yaml files with **$kubeletconf** , the remediation would read like: 

> edit the **kubelet service file /var/lib/kubelet/config.yaml** on each worker node

It's not technically accurate since that's not pointing to the real systemd service file, but it shows them the file that needs to be changed.

On the other hand, if replace the variable to use **$kubeletsvc**, it would read like:

> edit the kubelet service file **/etc/systemd/system/kubelet.service.d/10-kubeadm.conf** on each worker node

It's the correct systemd service file, but it doesn't accurately indicate the file that needs to be fixed.

robertojrojas:
I agree. I'll make the changes.

Issue #364, opened 2019-07-12
Add Unit Tests for the check/test.go:toNumeric function 


robertojrojas:
Add Unit Tests and refactor as needed, specially to remove the `os.Exit(1)` statements.

robertojrojas:
fixed by #401

PR #401, opened 2019-08-28
Adds Unit Tests for check/toNumeric

robertojrojas:
Fixes issue #364
Adds Unit Tests for check/toNumeric

### DIFF
--- a/check/test.go
+++ b/check/test.go
@@ -164,22 +164,22 @@ func compareOp(tCompareOp string, flagVal string, tCompareValue string) (string,
 
 	case "gt":
 		expectedResultPattern = "%s is greater than %s"
-		a, b := toNumeric(flagVal, tCompareValue)
+		a, b := toNumeric(flagVal, tCompareValue, newOsExit())
 		testResult = a > b
 
 	case "gte":
 		expectedResultPattern = "%s is greater or equal to %s"
-		a, b := toNumeric(flagVal, tCompareValue)
+		a, b := toNumeric(flagVal, tCompareValue, newOsExit())
 		testResult = a >= b
 
 	case "lt":
 		expectedResultPattern = "%s is lower than %s"
-		a, b := toNumeric(flagVal, tCompareValue)
+		a, b := toNumeric(flagVal, tCompareValue, newOsExit())
 		testResult = a < b
 
 	case "lte":
 		expectedResultPattern = "%s is lower or equal to %s"
-		a, b := toNumeric(flagVal, tCompareValue)
+		a, b := toNumeric(flagVal, tCompareValue, newOsExit())
 		testResult = a <= b
 
 	case "has":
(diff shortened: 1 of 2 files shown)


check/test.go
```diff
@@ -342,17 +342,36 @@ func (ts *tests) execute(s string) *testOutput {
...
+		oe = &osExit{}
+	}
+	return oe
+}
+
+func (eh *osExit) Exit(code int) {
+	os.Exit(code)
+}
+
+func toNumeric(a, b string, e exiter) (c, d int) {
```
danielpacak: I'm thinking whether the exiter abstraction is absolutely necessary at this level. 

I guess that testing `toNumeric` method was challenging due to `s.Exit` call and `fmt.Fprintf` to the std out. Both are modifying so called global state which, in general, we're trying to avoid in unit tests.

What if we've changed the signature to return an error as 3rd param and let the caller of `toNumeric` decide whether to `os.Exit` or not?

```go
func toNumeric(a, b string) (c, d int, err error) {
	c, err = strconv.Atoi(strings.TrimSpace(a))
	if err != nil {
		return 0, 0, fmt.Errorf("converting %s to numeric: %v", a, err)
	}
	d, err = strconv.Atoi(strings.TrimSpace(b))
	if err != nil {
		return 0, 0, fmt.Errorf("converting %s to numeric: %v", b, err)
	}

	return c, d, nil
}
```

By doing that you could write somehow lighter tests:

```go
_, _, err := toNumeric("NaN", 3)
assert.EqualError(t, err, "converting NaN to numeric: some underlying error")
_, _, err := toNumeric(5, "NaN")
assert.EqualError(t, err, "converting NaN to numeric: some underlying error")
```

I believe it's more idiomatic and gives you an opportunity to assert on error messages, which is very important for debugging and troubleshooting.

check/test_test.go
```diff
@@ -656,3 +656,45 @@ func TestCompareOp(t *testing.T) {
...
+	toNumeric("a", "b", me)
+	if me.expectedExitCode != me.passedExitCode {
+		t.Errorf("TestToNumeric - Expected Exit Code %d , but instead got %d", me.expectedExitCode, me.passedExitCode)
+	}
+
+	me = &mockExiter{
+		expectedExitCode: 1,
+	}
+
+	toNumeric("5", "b", me)
```
danielpacak: This looks to me as a perfect candidate for a [table driven test](https://github.com/golang/go/wiki/TableDrivenTests), where you could have an anonymous struct which args passed to `toNumeric` method and expected results and errors if any (see also my previous comment about adding 3rd arg to `toNumeric` to facilitate unit testing).

check/test.go
robertojrojas: I like your approach better than mine. 💯

check/test_test.go
robertojrojas: Agreed 💯

PR #413, opened 2019-09-02
README: Map CIS Benchmarks to releases

mrueg:
Let users know which release is the minimum requirement for a CIS
Benchmark.

Data source: the release, the first commit in the versioned folder was included.


lizrice:
Sorry we've left this for ages. The concern I have here is that it would encourage people to use old releases of kube-bench. Newer releases often include fixes that affect older benchmarks, and we would rather have people use the latest version wherever possible. Please re-open if you feel I'm missing something important here.

mrueg:
No I get you. The meaning of the table has changed a bit and is more dynamic. Thanks for getting back and closing!

PR #414, opened 2019-09-03
add snap component paths to default config

kwmonroe:
Rework #389 based on review.  Rather than including a snap-based config subdirectory, let's add snap component paths to the default config.  Without this, we fail the configuration file tests like this:

```
[INFO] 1.4 Configuration Files
[FAIL] 1.4.1 Ensure that the API server pod specification file permissions are set to 644 or more restrictive (Scored)
[FAIL] 1.4.2 Ensure that the API server pod specification file ownership is set to root:root (Scored)
[FAIL] 1.4.3 Ensure that the controller manager pod specification file permissions are set to 644 or more restrictive (Scored)
[FAIL] 1.4.4 Ensure that the controller manager pod specification file ownership is set to root:root (Scored)
[FAIL] 1.4.5 Ensure that the scheduler pod specification file permissions are set to 644 or more restrictive (Scored)
[FAIL] 1.4.6 Ensure that the scheduler pod specification file ownership is set to root:root (Scored)
```

With it, those same tests pass:

```
[INFO] 1.4 Configuration Files
[PASS] 1.4.1 Ensure that the API server pod specification file permissions are set to 644 or more restrictive (Scored)
[PASS] 1.4.2 Ensure that the API server pod specification file ownership is set to root:root (Scored)
[PASS] 1.4.3 Ensure that the controller manager pod specification file permissions are set to 644 or more restrictive (Scored)
[PASS] 1.4.4 Ensure that the controller manager pod specification file ownership is set to root:root (Scored)
[PASS] 1.4.5 Ensure that the scheduler pod specification file permissions are set to 644 or more restrictive (Scored)
[PASS] 1.4.6 Ensure that the scheduler pod specification file ownership is set to root:root (Scored)
```


robertojrojas:
@kwmonroe Thanks for your contribution! 
I'd like to perform some tests. What would be the easiest way to get a bare-metal snap-based k8s installation up and running?

kwmonroe:
@robertojrojas the *easiest* way to get a bare-metal snap-based k8s is to install the snaps on a vanilla ubuntu system, minimally configure them, and restart the various systemd services:

```bash
## install snaps
for i in docker etcd kube-apiserver kube-controller-manager kube-proxy kube-scheduler kubectl kubelet; do sudo snap install $i --classic; done

## configure snaps
sudo cp /snap/etcd/current/etcd.conf.yml.sample /var/snap/etcd/common/etcd.conf.yml
sudo snap set kube-apiserver cert-dir=/var/snap/kube-apiserver/current/tmp etcd-servers=http://127.0.0.1:2379
sudo snap set kube-controller-manager master="http://127.0.0.1:8080"
sudo snap set kube-scheduler master="http://127.0.0.1:8080"
sudo snap set kube-proxy master="http://127.0.0.1:8080"

## restart services
sudo systemctl restart snap.etcd.etcd
sudo systemctl restart snap.kube-apiserver.daemon
sudo systemctl restart snap.kube-controller-manager.daemon
sudo systemctl restart snap.kube-scheduler.daemon
sudo systemctl restart snap.kube-proxy.daemon
sudo systemctl restart snap.kubelet.daemon
```

At this point, you can install `kube-bench` from source and test the config in this PR:

```bash
## install go if needed
sudo snap install go --classic
mkdir ~/go
export GOPATH=~/go

## install kube-bench
go get github.com/aquasecurity/kube-bench
go get github.com/golang/dep/cmd/dep
cd $GOPATH/src/github.com/aquasecurity/kube-bench
$GOPATH/bin/dep ensure -vendor-only
go build -o kube-bench .

## run kube-bench
./kube-bench master
...
== Summary ==
16 checks PASS
48 checks FAIL
26 checks WARN
1 checks INFO

./kube-bench node
...
== Summary ==
4 checks PASS
17 checks FAIL
2 checks WARN
1 checks INFO

## checkout this PR to get snap-based config
git fetch origin pull/414/head:pr/414 && git checkout pr/414

## run kube-bench again to see the improvement when snap configs are considered
./kube-bench master
...
== Summary ==
24 checks PASS
40 checks FAIL
26 checks WARN
1 checks INFO

./kube-bench node
...
== Summary ==
8 checks PASS
13 checks FAIL
2 checks WARN
1 checks INFO

## remove the snaps if desired
for i in docker etcd kube-apiserver kube-controller-manager kube-proxy kube-scheduler kubectl kubelet; do sudo snap remove $i; done
```

As I noted in the beginning, the above is probably the *easiest* way to get a snap-based k8s cluster, but it certainly isn't a good one -- there are no certs, no addons (e.g. dns, dashboard, heapster, etc), and all services are talking insecurely over localhost.  However, it does let you have a look at the snap config files that are proposed in this PR.

For a real-world snap-based k8s cluster, there's [Charmed Kubernetes](https://ubuntu.com/kubernetes/docs), which is typically installed on public or private clouds as described [here](https://ubuntu.com/kubernetes/docs/quickstart). If you'd like, I can spin up a deployment for you to check out -- just point me at your public ssh key for access.

robertojrojas:
@kwmonroe thanks for the links. I was able go get a cluster following [quickstart](https://ubuntu.com/kubernetes/docs/quickstart). Do you know if there are many differences between a Cloud cluster installation and  private one using the quick start?

kwmonroe:
> Do you know if there are many differences between a Cloud cluster installation and private one using the quick start?

@robertojrojas there are some differences in a Charmed Kubernetes deployment on public clouds vs private clouds (openstack) vs private metal, but not many.  Things like:

- `cloud-provider` is set appropriately when deployed on a cloud/openstack/vsphere/etc
- kube-apiserver `allow-privileged` will be set if a gpu or ceph storage is detected
- kube-apiserver `auth*-webhook-config-file` will be set if keystone is detected

Most snap config will be consistent across all substrates -- the differences come up when features (ceph, keystone, etc) or platform-specifics (GPUs, cloud type, etc) are detected.

robertojrojas:
@kwmonroe I noticed, at least on my bare-metal installation of CDK, that the worker nodes uses these directories. are these the same on other environments?

- /root/cdk/kubelet/config.yaml
- /root/cdk/kubeproxyconfig
- /root/cdk/kubeconfig

kwmonroe:
@robertojrojas yes, those directories are used by Charmed Kubernetes across all environments (metal, cloud, lxc, etc).  When you execute something like `kubelet` on a worker with the kubelet snap installed, the binary is called with arguments from `/var/snap/kubelet/current/args` -- hence all the `args` files being included as `confs` in this PR.

For CDK, things like secrets and dynamic cluster config are set in `/root/cdk/<configs>` and those paths are referenced in the appropriate `/var/snap/<snap>/current/args` files. I'm intentionally not including the files in `/root/cdk` as `confs` here for a few reasons:

- the path feels too specific to be in the top-level default config
- i imagine not everyone runs kube-bench in an env with root access
- the `/root/cdk` configs seem to align more closely with the `1.13-json` version work done for EKS

Hope that helps explain the difference/linkage between `/var/snap/<snap>/current/args` and `/root/cdk/<configs>`, and why only the former are included in this PR.

lizrice:
I think this change as-is will improve the situation e.g. for the 1.4.x tests shown in the first comment, so I'd be prepared to take this PR. But, a couple of thoughts: 

* If kube-bench can't access files like /root/cdk/kubelet/config.yaml, it is not going to be able to check the kubelet configuration for the worker node tests 
* In some environments you have to run kube-bench as root simply because you need it to access some of the applicable configuration files 

Wdyt @kwmonroe?

kwmonroe:
@lizrice +1 on having the 1.4 tests work ootb.  I'm not too worried about access to files under `/root/cdk` because that's a CDK issue, and not the more generic snap issue that I'm hoping to solve here.  If you `snap install <some-k8s-snap>`, the config will come from `/var/snap/...`, which is readable without root, and that's what I want kube-bench to include in its various default search paths.

If someone *did* need kube-bench to read `/root/cdk/$foo`, I would expect them to mount the appropriate dirs in the container, or pass in a `--version` subdir with appropriate overrides knowing they have to run as root -- both those scenarios are outlined in the main readme, so i think we're covered there.

PR #424, opened 2019-09-22
usr-bin volume mount not required

saiyam1814:
usr-bin volume mount not required as using kubelet version in command


saiyam1814:
@lizrice

saiyam1814:
Hi @lizrice can you have a look at it ? Not sure whom else to tag

Issue #432, opened 2019-10-04
Check 1.1.33 doesn't check alternate flag and fails when NodeRestriction is on


mwwolters:
1.1.33 only checks the **--enable-admission-plugins** flag to see if NodeRestriction is enabled, but there is another flag **--admission-control** that is currently used to enable NodeRestriction. enable-admission-plugins is the [replacement](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#is-there-a-recommended-set-of-admission-controllers-to-use) for admission-control, but both should be checked for the time being.

lizrice:
It's not clear to me which version of Kubernetes retires `--admission-control` completely. 

We try to follow the CIS Benchmark, which doesn't include this alternative deprecated flag `--admission-control`. So if we take this change we're diverging from the benchmark and we're potentially encouraging people to use a deprecated flag.

robertojrojas:
Closing as per comments above.

PR #433, opened 2019-10-04
Changed 1.1.33 to also check admission-control flag

mwwolters:
Currently there are two ways to enable admission controllers, through **--enable-admission-plugins** and **--admission-control.** This change checks both of these.

#432 for context.


robertojrojas:
@mwwolters thanks for your contribution, but as explained in the comments [432](https://github.com/aquasecurity/kube-bench/issues/432#issuecomment-552934287)
we won't be taking this changes.

Issue #339, opened 2019-07-03
SIGSEGV: segmentation violation when trying to use a custom config


lucazz:
Hello everyone, 

I'm trying to override a couple of those checks on our gke clusters and am getting these errors when doing so: 
```
~ $ kubectl logs kube-bench-4kc5m
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x7d7768]

goroutine 1 [running]:
github.com/spf13/viper.(*Viper).find(0x0, 0x969617, 0xa, 0xa, 0x30)
        /go/pkg/mod/github.com/spf13/viper@v1.0.0/viper.go:865 +0x38
github.com/spf13/viper.(*Viper).Get(0x0, 0x969617, 0xa, 0xc0000dfbb8, 0x40ccad)
        /go/pkg/mod/github.com/spf13/viper@v1.0.0/viper.go:594 +0x7c
github.com/spf13/viper.(*Viper).GetStringSlice(0x0, 0x969617, 0xa, 0xc0000a01e0, 0x96797c, 0x6)
        /go/pkg/mod/github.com/spf13/viper@v1.0.0/viper.go:691 +0x3f
github.com/aquasecurity/kube-bench/cmd.getBinaries(0x0, 0x96797c, 0x6, 0x0)
        /go/src/github.com/aquasecurity/kube-bench/cmd/util.go:85 +0x66
github.com/aquasecurity/kube-bench/cmd.isMaster(0xc0000a14a0)
        /go/src/github.com/aquasecurity/kube-bench/cmd/common.go:237 +0x8d
github.com/aquasecurity/kube-bench/cmd.glob..func4(0xd85320, 0xdab9d8, 0x0, 0x0)
        /go/src/github.com/aquasecurity/kube-bench/cmd/root.go:60 +0x26
github.com/spf13/cobra.(*Command).execute(0xd85320, 0xc00001e270, 0x0, 0x0, 0xd85320, 0xc00001e270)
        /go/pkg/mod/github.com/spf13/cobra@v0.0.1/command.go:702 +0x289
github.com/spf13/cobra.(*Command).ExecuteC(0xd85320, 0xdab9d8, 0x0, 0x0)
        /go/pkg/mod/github.com/spf13/cobra@v0.0.1/command.go:783 +0x2ca
github.com/spf13/cobra.(*Command).Execute(...)
[...]

PR #434, opened 2019-10-04
Properly initialize viper library when checking for master components

simar7:
This PR includes:
* Properly initializes the viper library config when checking for master components.
* Adds some additional test coverage to help guard against future regressions.

Fixes: https://github.com/aquasecurity/kube-bench/issues/339

Signed-off-by: Simarpreet Singh <simar.singh@aquasec.com>


cmd/common.go
```diff
@@ -242,7 +242,10 @@ func loadConfig(nodetype check.NodeType) string {
 func isMaster() bool {
 	glog.V(2).Info("Checking if the current node is running master components")
 	masterConf := viper.Sub(string(check.MASTER))
-	components, err := getBinaries(masterConf)
+	if masterConf == nil {
```
robertojrojas: We should add a log statement indicating why isMaster is returning `false` in this situation. It'll help with troubleshooting later on. I think glog.V(2).Info(...) should be fine.

cmd/common_test.go
```diff
@@ -110,3 +113,135 @@ func TestNewRunFilter(t *testing.T) {
 	})
 
 }
+
```
robertojrojas: I like the fact the testing using file data. I think it'll be better to have the contents to be in its own files, like `master_only.yaml`, `node_only.yaml`, `master_node.yaml`. The benefit I see of having external files is that that it'll be easier to check of file format and modifications as these files change later one.

cmd/common.go
simar7: Added.

cmd/common_test.go
simar7: Addressed.

Issue #395, opened 2019-08-27
Indicate when test is expected to be performed manually


lizrice:
At the moment we [WARN] when a user is expected to run a test manually, but they may not realise this. Suggest that we add [Manual test] to the remediation output for any test marked as "manual".

lizrice:
fixed by #435

PR #435, opened 2019-10-07
Add [Manual test] to remediation in all the manual tests

yoavrotems:


yoavrotems:
Relating to https://github.com/aquasecurity/kube-bench/issues/395
But I believe there are some tests here that we need to reconsider if it should be a manual test

Issue #177, opened 2018-11-08
1.1.37 should also check that --audit-policy-file is set


lizrice:
See discussion in PR #173

ttousai:
@lizrice

To resolve this issue, I think kube-bench needs to support sub-checks under checks and combining the results of each sub-check with operators like `and` or `or`.

For example, we could have check 1.1.37a and 1.1.37b then combine them with the comparison operators (for example and/or) to get the result for 1.1.37.
wdyt?
WDYT?

simar7:
This has been fixed here: https://github.com/aquasecurity/kube-bench/pull/440

PR #440, opened 2019-10-11
master.yaml: Add --audit-policy-file check for 1.1.37.

simar7:
Addresses: https://github.com/aquasecurity/kube-bench/issues/177

Signed-off-by: Simarpreet Singh <simar@linux.com>


simar7:
I believe the massive diff in 1.13 master.yaml is because it was checked in with Windows style line endings.

simar7:
> Let's keep the newline as they originally were in **`1.13/master.yaml`**.
> This is going to make it easy to keep track and see the code differences across changes.
> Later on, we'll migrate all files to be the same newline.

I've fixed that in the recent commit.

Issue #441, opened 2019-10-11
Check 1.3.3 fails when --use-service-account-credentials is present but not explicitly =true


mwwolters:
The setting controlled by the flag --use-service-account-credentials can be enabled either through **--use-service-account-credentials=true** or **--use-service-account-credentials**, but this check fails for the second case.

robertojrojas:
@mwwolters Thanks for your contribution!
Fixed by #442 
Closing.

PR #442, opened 2019-10-11
Change 1.3.3 to pass for --use-service-account-credentials=true and --use-service-account-credentials

mwwolters:
The setting controlled by the flag --use-service-account-credentials can be enabled either through --use-service-account-credentials=true or --use-service-account-credentials, but this check fails for the second case. This change checks for that case.

#441


cfg/1.11/master.yaml
```diff
@@ -544,12 +544,18 @@ groups:
...
       - flag: "--enable-admission-plugins"
         compare:
           op: has
           value: "NodeRestriction"
         set: true
+      - flag: "--admission-control"
+        compare:
+          op: has
+          value: "NodeRestriction"
+        set: true
```
lizrice: I think this change is something else

cfg/1.13/master.yaml
```diff
@@ -548,12 +548,18 @@ groups:
...
       - flag: "--enable-admission-plugins"
         compare:
           op: has
           value: "NodeRestriction"
         set: true
+      - flag: "--admission-control"
+        compare:
+          op: has
+          value: "NodeRestriction"
+        set: true
```
lizrice: Also something else, I think

cfg/1.11/master.yaml
mwwolters: Whoops, you're right. I think it should be fixed now.

cfg/1.13/master.yaml
mwwolters: Same as above.

PR #443, opened 2019-10-11
Update config.yaml

shivamkainth:
changed config file location


lizrice:
No, this change is incorrect

PR #445, opened 2019-10-11
Added license section and go report card badge

sumitkharche:


lizrice:
License should only be documented in one place, the LICENSE file

PR #447, opened 2019-10-13
Update README.md

sakshamtaneja21:
there are so many errors in this file 
if you say ill correct all
also is it kube or Kube


lizrice:
It's lower-case kube-bench, and many of the other suggestions here are grammatically incorrect

sakshamtaneja21:
ok

PR #465, opened 2019-10-19
Updated readme

nishu88:
Added right punctuations at important spots, And added License details at the end of README.md to match the layout of other repositories of aquasecurity's readme file.
Ex- https://github.com/aquasecurity/trivy/blob/master/README.md
Capture


lizrice:
License should only be documented in one place the LICENSE file

Issue #472, opened 2019-10-21
Fix grammatical issues in CONTRIBUTING.md


mathagician:

DhruvTrehan:
[https://github.com/aquasecurity/kube-bench/pull/490](url)

robertojrojas:
The documentation as it stands is correct, so I'll be closing this.

PR #473, opened 2019-10-21
Fix grammatical issues in CONTRIBUTING.md

mathagician:
This PR improves the content of CONTRIBUTING.md file from a grammar perspective without altering the meaning of the original text.

Reference: Issue #472


lizrice:
The typo is already fixed in #471 and the other changes are unnecessary, less clear, or change the meaning incorrectly

PR #478, opened 2019-10-21
Fix grammatical mistakes

superkd37:


simar7:
The documentation is correctly formatted as it stands.

Issue #269, opened 2019-04-24
RFC: change config numbering to reflect the spec rather than the Kubernetes version


lizrice:
Currently the config files are numbered according to the earliest Kubernetes version for which the config is valid. As described in [this table in the README](https://github.com/aquasecurity/kube-bench#cis-kubernetes-benchmark-support), in most cases the CIS spec is valid for more than Kubernetes version, and this causes some confusion. 

We also now have config files for Openshift, and for JSON-specific versions of the files. I am wondering if it would be clearer to rename them something like this: 

| kube-bench config | Spec |
|:-----|:-----|
| cis-1.0.0 | CIS Kubernetes Benchmark 1.0.0 |
| cis-1.1.0 | CIS Kubernetes Benchmark 1.1.0 | 
| ...etc... | |
| rh-0.7 | Red Hat Openshift Hardening Guide 0.7 |
| cis-1.4.0-json | CIS Kubernetes Benchmark, for platforms configured using json / yaml |

Auto-detection should still work i.e. we'd include intelligence in `kube-bench` to know which CIS spec version is valid for which Kubernetes version. 

Comments welcome! 

cc @jerbia, @liranaqua

lizrice:
@jerbia you are supportive of this, right?

lizrice:
We would continue to support `--version` flag, which specifies the Kubernetes version. User can specify, say, `--version 1.12` and kube-bench should map this to the right CIS version. 

We would also add a flag `--benchmark`, which specifies the config directory, e.g. `--benchmark cis-1.1.0` which maps directly to the `cfg/cis-1.10` directory. 

It would be an error to specify both `--version` and `--benchmark` flags.

robertojrojas:
fixed by #511

PR #480, opened 2019-10-21
Fixes Issue #269: Changes config numbering to use CIS versions

robertojrojas:
Fixes Issue #269: Changes config numbering to use CIS versions.

Introduced the flag **`--benchmark`**, which specifies the config directory, e.g. --benchmark **cis-1.4** which maps directly to the cfg/**cis-1.4** directory.


cmd/common.go
```diff
@@ -27,6 +27,16 @@ import (
...
 
+var kubeToCISMap = map[string]string{
+	"1.11": "cis-1.3.0",
+	"1.12": "cis-1.3.0",
+	"1.13": "cis-1.4.1",
+	"1.14": "cis-1.4.1",
+	"1.15": "cis-1.4.1",
+	"1.16": "cis-1.4.1",
+	"1.17": "cis-1.4.1",
+}
```
lizrice: Let's keep this future-resistant. If we kept the previous version of `decrementVersion` which decrements the kubernetes version, we could decrement until we find a version that has an entry in this table. So, something like

```
cisVersion, ok = kubeToCISmap[kubeVersion]
if !ok {
    for ; kubeVersion = decrementVersion(kubeVersion); (kubeVersion != 1.11 && !ok) { 
        cisVersion, ok = kubeToCISmap[kubeVersion]
   }
}

if !ok {
   we didn't find a version so error
}
```

cmd/common.go
```diff
@@ -27,6 +27,16 @@ import (
 	"github.com/spf13/viper"
 )
 
+var kubeToCISMap = map[string]string{
+	"1.11": "cis-1.3.0",
```
lizrice: I'm wondering whether we should omit the point release part of the spec number, so: cis-1.3, cis-1.4, cis-1.5 etc. Then the user doesn't need to know or care exactly when we update from, say, 1.4.1 to 1.4.2 if the CIS release an error-correcting update.

cmd/util_test.go
```diff
@@ -408,31 +408,27 @@ func TestGetConfigFilePath(t *testing.T) {
...
-		{runningVersion: "1.8", succeed: true, exp: d},
-		{runningVersion: "1.9", succeed: true, exp: d},
-		{runningVersion: "1.10", succeed: true, exp: d},
-		{runningVersion: "1.1", succeed: false},
-		{specifiedVersion: "1.8", succeed: true, exp: d},
-		{specifiedVersion: "1.9", succeed: false},
-		{specifiedVersion: "1.10", succeed: false},
+		{benchmarkVersion: "cis-1.4.1", specifiedVersion: true, succeed: true, exp: d},
+		{benchmarkVersion: "cis-1.5.0", specifiedVersion: false, succeed: true, exp: d},
+		{benchmarkVersion: "1.1", succeed: false},
```
lizrice: I think we should retain the tests that cover the user specifying the kubeVersion. Not sure we haven't also lost cases for when the user doesn't specify the version (no benchmark version and no kube version) and we have to work it out by getting the running version?

cmd/root.go
```diff
@@ -113,6 +114,7 @@ func init() {
 	RootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is ./cfg/config.yaml)")
 	RootCmd.PersistentFlags().StringVarP(&cfgDir, "config-dir", "D", "./cfg/", "config directory")
 	RootCmd.PersistentFlags().StringVar(&kubeVersion, "version", "", "Manually specify Kubernetes version, automatically detected if unset")
+	RootCmd.PersistentFlags().StringVar(&benchmarkVersion, "benchmark", "", "Manually specify CIS benchmark version, automatically detected if unset")
```
lizrice: I think we said we were going to make it an error if the user tries to set both kubeVersion and benchmarkVersion, right?

robertojrojas: 💯

cmd/util_test.go
robertojrojas: 💯

cmd/common.go
```diff
@@ -209,15 +219,23 @@ func loadConfig(nodetype check.NodeType) string {
 		file = nodeFile
 	}
 
-	runningVersion := ""
-	if kubeVersion == "" {
-		runningVersion, err = getKubeVersion()
-		if err != nil {
-			exitWithError(fmt.Errorf("Version check failed: %s\nAlternatively, you can specify the version with --version", err))
+	specificBenchmarkVersion := isNotEmpty(benchmarkVersion)
+	if !specificBenchmarkVersion {
```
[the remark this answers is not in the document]
robertojrojas: 💯

cmd/util.go
```diff
@@ -344,3 +371,11 @@ func makeSubstitutions(s string, ext string, m map[string]string) string {
 
 	return s
 }
+
+func isNotEmpty(str string) bool {
+	return !isEmpty(str)
+}
```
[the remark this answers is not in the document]
robertojrojas: Yes, I agree we don't need the `isNotEmpty` function, but I do find  `isEmpty(str)` easier to read than `str == ""`.
Viper probably doesn't give us a whitespace-only parameter, but I was thinking we could use `isEmpty(...)` later down the road.

cmd/common.go
robertojrojas: Yes, I like this idea. Users shouldn't care about minor updates. Another benefit is less config directories.
I wonder if it makes sense to keep the kubeVersion to CIS version mapping in a `version_mapping.yaml` file.
This file could also include the **`defaultKubeVersion`** 
I think it would make it easier to update the versions down the road.

cmd/common.go
robertojrojas: I used your idea as a starting point and made changes. wdyt?

cmd/common.go
```diff
@@ -237,6 +238,64 @@ func loadConfig(nodetype check.NodeType) string {
...
+
+func loadVersionMapping(v *viper.Viper) (map[string]string, error) {
+	versionMappingConf := v.Sub(string(versionMapping))
+	if versionMappingConf == nil {
+		return nil, fmt.Errorf("config file is missing %s section", versionMapping)
+	}
+	kubeToCISMap := versionMappingConf.GetStringMapString(kubersionCISMapping)
+	if kubeToCISMap == nil || (len(kubeToCISMap) == 0) {
+		return nil, fmt.Errorf("config file is missing %s.%s section", versionMapping, kubersionCISMapping)
+	}
```
lizrice: Per comment above, if we just had `version_mapping` this function could be half the length

cmd/common.go
```diff
@@ -237,6 +238,64 @@ func loadConfig(nodetype check.NodeType) string {
...
+		if err != nil {
+			return "", err
+		}
+
+		benchmarkVersion = mapToCISVersion(kubeToCISMap, kubeVersion)
+		if isEmpty(benchmarkVersion) {
+			// unable to find a match for kubeVersion
+			return "", fmt.Errorf("Unable to find a matching CIS Version match for kubernetes version: %s", kubeVersion)
+		}
+		glog.V(2).Info(fmt.Sprintf("Mapped Kubernetes version: %s to CIS version: %s", kubeVersion, benchmarkVersion))
```
lizrice:
```suggestion
		glog.V(2).Info(fmt.Sprintf("Mapped Kubernetes version: %s to benchmark: %s", kubeVersion, benchmarkVersion))
```

to allow for the possibility of benchmarks that aren't CIS, like Red Hat

cmd/common.go
```diff
@@ -237,6 +238,64 @@ func loadConfig(nodetype check.NodeType) string {
...
+		kubeToCISMap, err := loadVersionMapping(v)
+		if err != nil {
+			return "", err
+		}
+
+		benchmarkVersion = mapToCISVersion(kubeToCISMap, kubeVersion)
+		if isEmpty(benchmarkVersion) {
+			// unable to find a match for kubeVersion
+			return "", fmt.Errorf("Unable to find a matching CIS Version match for kubernetes version: %s", kubeVersion)
+		}
```
lizrice: This is really _not_ a big deal, but it's a bit odd that one of these functions returns an error, and the next uses empty return value to infer an error. It would be slightly easier to read this code if you settled on one approach or the other!

cmd/common.go
```diff
@@ -237,6 +238,64 @@ func loadConfig(nodetype check.NodeType) string {
...
+
+		kubeToCISMap, err := loadVersionMapping(v)
+		if err != nil {
+			return "", err
+		}
+
+		benchmarkVersion = mapToCISVersion(kubeToCISMap, kubeVersion)
+		if isEmpty(benchmarkVersion) {
+			// unable to find a match for kubeVersion
+			return "", fmt.Errorf("Unable to find a matching CIS Version match for kubernetes version: %s", kubeVersion)
```
lizrice:
```suggestion
			return "", fmt.Errorf("Unable to find a matching benchmark for kubernetes version: %s", kubeVersion)
```

cmd/common_test.go
```diff
@@ -165,3 +167,131 @@ func TestIsMaster(t *testing.T) {
...
+	cases := []struct {
+		kubeVersion string
+		succeed     bool
+		exp         string
+	}{
+		{kubeVersion: "1.11", succeed: true, exp: "cis-1.3"},
+		{kubeVersion: "1.12", succeed: true, exp: "cis-1.3"},
+		{kubeVersion: "1.13", succeed: true, exp: "cis-1.4"},
+		{kubeVersion: "1.16", succeed: true, exp: "cis-1.4"},
+		{kubeVersion: "unknown", succeed: false, exp: ""},
```
lizrice: Maybe add a test for a version lower than we support e.g. 1.10, which should fail

cmd/common_test.go
```diff
@@ -165,3 +167,131 @@ func TestIsMaster(t *testing.T) {
...
+	cases := []struct {
+		n                string
+		kubeVersion      string
+		benchmarkVersion string
+		v                *viper.Viper
+		exp              string
+		succeed          bool
+	}{
+		{n: "both versions", kubeVersion: "1.11", benchmarkVersion: "cis-1.3", exp: "cis-1.3", v: viper.New(), succeed: false},
+		{n: "no version", kubeVersion: "", benchmarkVersion: "", v: viperWithData, exp: "cis-1.3", succeed: true},
```
lizrice: I'm not sure about this "no version" test. If the user doesn't specify either kubeVersion or benchmarkVersion, we should be using `getKubeVersion()` to find the running version of kubernetes, and failing if we can't get it. In unit testing we shouldn't be getting a running version of kubernetes, should we? (Unless that is mocked out and I am failing to see that)

cmd/common.go
```diff
@@ -237,6 +238,64 @@ func loadConfig(nodetype check.NodeType) string {
 	return filepath.Join(path, file)
 }
 
+func mapToCISVersion(kubeToCISMap map[string]string, kv string) string {
```
lizrice:
```suggestion
func mapToBenchmark(kubeToBenchmark map[string]string, kv string) string {
```

Avoiding saying "CIS" to allow for things like Red Hat's document

cmd/common.go
```diff
@@ -237,6 +238,64 @@ func loadConfig(nodetype check.NodeType) string {
...
+	cisVersion, found := kubeToCISMap[kv]
```
lizrice:
```suggestion
	benchmark, found := kubeToBenchmark[kv]
```

and similar throughout

cmd/common.go
```diff
@@ -237,6 +238,64 @@ func loadConfig(nodetype check.NodeType) string {
...
+	for !found && (kv != defaultKubeVersion && !isEmpty(kv)) {
+		kv = decrementVersion(kv)
+		cisVersion, found = kubeToCISMap[kv]
+		glog.V(2).Info(fmt.Sprintf("mapToCISVersion for cisVersion: %q found: %t\n", cisVersion, found))
+	}
+
+	if !found {
+		glog.V(2).Info(fmt.Sprintf("mapToCISVersion unable to find a match for: %q", kv))
```
lizrice:
```suggestion
		glog.V(1).Info(fmt.Sprintf("mapToCISVersion unable to find a match for: %q", kv))
```
Since it's an error case I would make this log at level 1

cmd/common.go
```diff
@@ -237,6 +238,64 @@ func loadConfig(nodetype check.NodeType) string {
 	return filepath.Join(path, file)
 }
 
+func mapToCISVersion(kubeToCISMap map[string]string, kv string) string {
+	cisVersion, found := kubeToCISMap[kv]
+	for !found && (kv != defaultKubeVersion && !isEmpty(kv)) {
+		kv = decrementVersion(kv)
+		cisVersion, found = kubeToCISMap[kv]
+		glog.V(2).Info(fmt.Sprintf("mapToCISVersion for cisVersion: %q found: %t\n", cisVersion, found))
```
lizrice:
```suggestion
		glog.V(3).Info(fmt.Sprintf("mapToCISVersion for cisVersion: %q found: %t\n", cisVersion, found))
```

cmd/util_test.go
```diff
@@ -408,37 +408,66 @@ func TestGetConfigFilePath(t *testing.T) {
...
 	if err != nil {
 		t.Fatalf("Failed to create temp file")
 	}
 	ioutil.WriteFile(filepath.Join(d, "master.yaml"), []byte("hello world"), 0666)
 
 	cases := []struct {
-		specifiedVersion string
-		runningVersion   string
+		benchmarkVersion string
+		specifiedVersion bool
```
lizrice: I don't think this specifiedVersion field is used?

cmd/common_test.go
robertojrojas: yeah that's a bug. I'll fix that.

robertojrojas:
I'll work in changes to the REAME now.
Yes, testing is always good. :)

robertojrojas:
Fixed with #511

PR #484, opened 2019-10-23
Update README.md

Iltwats:


simar7:
I don't think this adds much value, the documentation remains correct as it stands.

PR #495, opened 2019-10-26
Adding a section for Azure Kubernetes Service

sauryadas:
steps to run kube bench on AKS worker nodes

@andyzhangx FYI


README.md
```diff
@@ -144,6 +145,31 @@ To run the tests on the master node, the pod needs to be scheduled on that node.
...
+
+1. config KUBECONFIG on the agent node:
+```
+sudo chmod a+r /etc/kubernetes/certs/client.key
+export KUBECONFIG=/var/lib/kubelet/kubeconfig
+kubectl get no  # to check whether kubeconfig works
+```
+4. Run CIS benchmark to view results:
+```
+docker run --rm -v `pwd`:/host andyzhangx/kube-bench:latest install
```
lizrice: shouldn't this be `aquasecurity/kube-bench:latest`?

sauryadas: Thanks @lizrice .. We had to make some specific changes to the kube-bench code . Should we publish an image with an Azure tag instead ? kube-bench:azure

@andyzhangx

lizrice: Let's see if we can find a way to incorporate the changes into this repo - can you share what needed to change?

sauryadas: @andyzhangx Can you please share the diff?

andyzhangx: the code change is here: https://github.com/andyzhangx/kube-bench/commit/5f5ee060445b762ae93945a8288e7de0c20caf9d
@lizrice

lizrice:
We could handle this in the main codebase by adding the ability to have alternative locations for kubeconfig files (which should be easy to do since we do it for lots of other files)

lizrice:
Thanks for pointing us to the code change. We now have `/var/lib/kubelet/kubeconfig` listed as options for kubeconfig for both kubelet and proxy, so kube-bench will now look for the file iin this location. This is the equivalent of what you have, but done in a slightly different way. @robertojrojas has been testing on AKS to make sure this works. Would be great if you can also confirm that it works for you @sauryadas @andyzhangx. 

If you can use the aquasecurity/kube-bench:latest image name we can merge this change

andyzhangx:
> Thanks for pointing us to the code change. We now have `/var/lib/kubelet/kubeconfig` listed as options for kubeconfig for both kubelet and proxy, so kube-bench will now look for the file iin this location. This is the equivalent of what you have, but done in a slightly different way. @robertojrojas has been testing on AKS to make sure this works. Would be great if you can also confirm that it works for you @sauryadas @andyzhangx.
> 
> If you can use the aquasecurity/kube-bench:latest image name we can merge this change

@lizrice sorry for the late reply, I have confirmed that we could run the CIS benchmark test with `aquasecurity/kube-bench:latest` image, now the step is quite simpler:
```
kubectl-enter {node-name}
docker run --rm -v `pwd`:/host aquasec/kube-bench:latest install
./kube-bench node
```

@sauryadas would you change the PR? Thanks.

sauryadas:
Updates PR @andyzhangx  @lizrice Thanks for the feedback ...

README.md
```diff
@@ -177,6 +178,31 @@ To run the tests on the master node, the pod needs to be scheduled on that node.
...
+1. Create an AKS cluster(e.g. 1.13.7) with RBAC enabled, otherwise there would be 4 failures
+
+1. Use the [kubectl-enter plugin] (https://github.com/kvaps/kubectl-enter) to shell into a node 
+`
+kubectl-enter {node-name}
+` 
+or ssh to one agent node
+could open nsg 22 port and assign a public ip for one agent node (only for testing purpose)
+
+1. config KUBECONFIG on the agent node:
```
andyzhangx: we don't need this step now, only need following steps:
```
kubectl-enter {node-name}
docker run --rm -v `pwd`:/host aquasec/kube-bench:latest install
./kube-bench node
```

sauryadas: that is why i say OR ssh .. if somehow kubectl-enter tool breaks there is another way to get to the node..

andyzhangx: I mean step 3, it's not necessary now:
```
config KUBECONFIG on the agent node:

sudo chmod a+r /etc/kubernetes/certs/client.key
export KUBECONFIG=/var/lib/kubelet/kubeconfig
kubectl get no  # to check whether kubeconfig works
```

sauryadas: ok.. removed now

sauryadas:
@lizrice please review when you get a chance

sauryadas:
@lizrice gentle reminder

Issue #331, opened 2019-06-28
need to differentiate failing to find kubectl/kubelet from other executables


lizrice:
Depending on the environment, failing to find the kubelet / kubectl might be resolved by specifying a version explicitly, or by specifying a different PATH. This is a different category of error from failing to find the control plane components. In these situations it may be that the config needs modifying to specify a different binary to look for. 

We should have different error messages in these two situations so that people are less likely to report them as the same issue (as in #328). We should also have kube-bench give better advice on how to modify the config to try to solve problems.

robertojrojas:
Fixed by #497

PR #497, opened 2019-10-27
Fixes Issue #331: Changes the Error Message When Programs are Missing  

robertojrojas:
Fixes Issue #331.

- When **kubectl** and **kubelet** are missing, the following error is displayed:

  ``` 
   Version check failed:  
    Unable to find the programs kubectl or kubelet in the PATH.
    These programs are used to determine version of kubernetes.
    Make sure the /usr/bin directory is mapped to the container,
    either using the job.yaml file used, or docker command.

    For job.yaml:
    ...
    - name: usr-bin
       mountPath: /usr/bin
    ...

    For docker command:
         docker -v $(which kubectl):/usr/bin/kubectl ....

    Alternatively, you can specify the version with --version
         kube-bench --version <VERSION> ...


- When component programs are missing, the following error is displayed:

   ```
   Unable to find control plane programs (apiserver, controllermanager, etc...)
   These program names are provided in the config.yaml, section master.apiserver.bins
   The following control plane programs for the component 'apiserver' have been search,
    but none of them have been found:
	  - kube-apiserver
	  - hyperkube apiserver
	  - hyperkube kube-apiserver
	  - apiserver
   ```

### DIFF
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -35,23 +35,7 @@ var TypeMap = map[string][]string{
 	"config":     []string{"confs", "defaultconf"},
 }
 
-var errMissingKubectlKubelet = fmt.Errorf(`Unable to find the programs kubectl or kubelet in the PATH.
-These programs are used to determine which version of Kubernetes is running.
-Make sure the /usr/bin directory is mapped to the container, 
-either in the job.yaml file, or Docker command.
-
-For job.yaml:
-...
-- name: usr-bin
-  mountPath: /usr/bin
-...
-
-For docker command:
-   docker -v $(which kubectl):/usr/bin/kubectl ....
-
-Alternatively, you can specify the version with --version
-   kube-bench --version <VERSION> ...
-`)
+var errMissingKubectlKubelet = fmt.Errorf(`unable to find the programs kubectl or kubelet in the PATH`)
 
 func init() {
 	psFunc = ps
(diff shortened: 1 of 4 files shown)


cmd/util.go
```diff
@@ -27,14 +27,32 @@ var (
...
-var getBinariesFunc func(*viper.Viper) (map[string]string, error)
+var getBinariesFunc func(*viper.Viper, check.NodeType) (map[string]string, error)
 var TypeMap = map[string][]string{
 	"ca":         []string{"cafile", "defaultcafile"},
 	"kubeconfig": []string{"kubeconfig", "defaultkubeconfig"},
 	"service":    []string{"svc", "defaultsvc"},
 	"config":     []string{"confs", "defaultconf"},
 }
 
+var errMissingKubectlKubelet = fmt.Errorf(`Unable to find the programs kubectl or kubelet in the PATH.
```
lizrice: I think golint will complain about these as error messages (which by convention are not capitalized and aren't supposed to be full sentences). Perhaps there should be two things: 

- a short error message (e.g. "unable to find %s" in PATH")
- a more helpful human-readable text message which gets printed to stderr by a top-level function

cmd/util.go
```diff
@@ -344,3 +362,30 @@ func makeSubstitutions(s string, ext string, m map[string]string) string {
 
 	return s
 }
+
+func buildComponentMissingErrorMessage(nodetype check.NodeType, component string, bins []string) error {
```
lizrice: Same comment as before about error messages - the type this returns could be string rather than error

cmd/util.go
```diff
@@ -27,14 +27,16 @@ var (
...
-var getBinariesFunc func(*viper.Viper) (map[string]string, error)
+var getBinariesFunc func(*viper.Viper, check.NodeType) (map[string]string, error)
 var TypeMap = map[string][]string{
 	"ca":         []string{"cafile", "defaultcafile"},
 	"kubeconfig": []string{"kubeconfig", "defaultkubeconfig"},
 	"service":    []string{"svc", "defaultsvc"},
 	"config":     []string{"confs", "defaultconf"},
 }
 
+var errMissingKubectlKubelet = fmt.Errorf(`unable to find the programs kubectl or kubelet in the PATH`)
```
lizrice: It would be clearer to have this as an error inline in the code, it doesn't need to be a variable

Issue #498, opened 2019-10-27
Run kube-bench image as non-root user


s-nirali:
This update is in view of following security best practices.

lizrice:
I don't think this is going to work on all platforms as there are some cases where it needs root privilege to look at certain files. @robertojrojas wdyt?

robertojrojas:
I think this might not work in cases where the code needs to read certain file, **/var/lib/kubelet/config.yaml** is one example that comes to mind.

s-nirali:
Hello @lizrice , @robertojrojas . I've tested the container created using a non-root user by mounting the /var and /etc directory . All contents of the directory are accessible inside to the non-root user. The results of running kube-bench with or without the root user are also the same. However, the "install" command in the entrypoint script does require root privileges to install the binary in the host. For that, I updated the makefile in the PR with the following usage command.

```
docker run --rm -v `pwd`:/host -u 0:0 aquasec/kube-bench:latest install
```

We've been using kube-bench in our organization and it's been a really great experience so far. Thank you for developing this 😃  . However, recently, I was asked to make this change which is why I raised an issue and a PR here. I'm not aware about any specific checks that require root to run. If there are, it would be great to have them documented so I can show it to my reviewers. It'll also be helpful if someone could thoroughly test the PR for any breaking checks or other issues that I might've missed.

leodotcloud:
It doesn't matter who the user is inside the container, if the docker run command is invoked as root.

lizrice:
It may be possible to run kube-bench as a non-root user on some platforms, but in general we believe that it will need to be root (on the host) to have correct access to certain config files, and to produce correct results. For that reason we are not proposing to pursue this issue. 

Note that the associated PR #499 sets up the image as a non-root user, but then overrides the user with -u 0:0 so that it does actually run the container as root.

PR #499, opened 2019-10-27
Run image as non-privileged user

s-nirali:
Resolves https://github.com/aquasecurity/kube-bench/issues/498

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 language: go
 
 services:
-  - docker 
+  - docker
 
 notifications:
   email: false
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -156,9 +156,12 @@ There are two significant differences on EKS:
...
 
 This command copies the kube-bench binary and configuration files to your host from the Docker container:
-** binaries compiled for linux-x86-64 only (so they won't run on macOS or Windows) **
+> **Note:**
+> - Binaries compiled for linux-x86-64 only (so they won't run on macOS or Windows).
+> - To install, the container must be run as root user.
+
 ```
-docker run --rm -v `pwd`:/host aquasec/kube-bench:latest install
+docker run --rm -v `pwd`:/host -u 0:0 aquasec/kube-bench:latest install
```
lizrice: As above, the -u 0:0 option explicitly runs the container as root!

.travis.yml
```diff
@@ -15,12 +15,12 @@ before_install:
 
 script:
   - GO111MODULE=on go test ./...
-  - docker build --tag kube-bench .
-  - docker run -v `pwd`:/host kube-bench install 
+  - docker build  --tag kube-bench .
+  - docker run -v `pwd`:/host -u 0:0 kube-bench install
```
lizrice: The -u 0:0 explicitly sets the user to root, overriding what is set in the image

README.md
s-nirali: This is necessary only to install the binary on the host. If it is run within the container with host mounted volumes, then it can run with non-root privileges.

PR #500, opened 2019-10-28
Remove unwanted hooks/build file

s-nirali:
Contents of the hooks/build file are already present in the makefile for
the `build-docker` target.


lizrice:
I think the hooks/build  is used by the automated build on docker hub

PR #505, opened 2019-10-30
Add helm chart using DaemonSet

s-nirali:
Resolves https://github.com/aquasecurity/kube-bench/issues/225

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 language: go
 
 services:
-  - docker 
+  - docker
 
 notifications:
   email: false
@@ -13,15 +13,21 @@ before_install:
   - gem install --no-ri --no-rdoc fpm
   - go get -t -v ./...
 
+install:
+  - wget ${HELM_URL}/${HELM_TGZ}
+  - tar -zxvf ${HELM_TGZ}
+  - PATH=`pwd`/linux-amd64/:$PATH
+
 script:
   - GO111MODULE=on go test ./...
   - IMAGE_NAME=kube-bench make build-docker
-  - docker run -v `pwd`:/host kube-bench install 
+  - docker run -v `pwd`:/host kube-bench install
   - test -d cfg
   - test -f kube-bench
   - make tests
   - make integration-tests
-  
+  - helm lint deploy/kube-bench/
+
 after_success:
   - bash <(curl -s https://codecov.io/bash)
 deploy:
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -33,6 +39,8 @@ deploy:
...
+    - secure: mb8AYZKDo6hkKN+2F9ldXcw27Yn2AfxpXvKlD8GD7NdGOI+TaiSFbE0I+qqTa/1DqcRekCQwqN7OG/17s9JDkgzUXYuYUGlVUOM4WbeJoSlzJFIOh9r9R/JddluYJohypgkE20IBHIrEHq5sY0Nn1Pl9WgSQFaVcQjxkX009AOuVjN0o5HcoXsb5hAzvHrpoSPkcSSqq7VWab60TgUttVaRlZSGwGdSYQEqk5TdO0hWHuXyxaaEPybgFIyZLLbxPS4JmMz8n3Sngetpw9Jgc+V9Fc7wKXpjvZZ33SpArG5p5ZFFu2YQOXFLZth9qtQOjduQ2gU1kHN6WjWnJ8QX2s8vmU38Tk19kd5i+mz9dvc87IdBvmTIqVYSpM6AAYa2osBGP3f97Rj2S68lTad4ecSVyHdsjz56vdE3ZH4wskswmogbKkVdvO4biPHxT6odszBxYLEJuRJyZ7ckXd52MCzqAUPrw7YUuH8N1mLIlf7V5bW5R+q4DlKw774zxnHiWrymXGvlINSrB0qxBn8Fii6ib+Pacl3PuqSumCcgIHlVjqrzIXaqcTMn2/ABZYC99mralGvwA/EgNa8CBKB5evMCEwWa5Ntvcs2I2DFcO5Q2WzN4H0YScyAzzCzK7/3hWJE/rUIJntwiSXkV3MSa1yxWSGGH8F1lcz+lzgTBm/MU=
+    - HELM_URL=https://get.helm.sh
+    - HELM_TGZ=helm-v3.0.0-rc.1-linux-amd64.tar.gz
```
itaysk: can you please update to the recent version?

deploy/kube-bench/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for Cuber CIS Benchmarks component using kube-bench
```
itaysk: "Cuber" is typo right? Not sure if cyber or kuber*

deploy/kube-bench/Chart.yaml
```diff
@@ -0,0 +1,5 @@
...
+name: kube-bench
+version: 0.1.1
```
itaysk: what's this version? This field should be the chart version so let's use 0.1.0

deploy/kube-bench/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
```
itaysk: this should be the version of kube-bench being installed

deploy/kube-bench/templates/daemonset.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
```
itaysk: we should include the chart labels here (`kube-bench.labels` helper). look at the generated templates when you run `helm create` for reference how it should look like.

deploy/kube-bench/templates/daemonset.yaml
```diff
@@ -0,0 +1,40 @@
...
+    metadata:
+      labels:
+        app: kube-bench
+      name: kube-bench
+    spec:
+      hostPID: true
+      containers:
+        - name: kube-bench
+          image: {{ .Values.image.repository }}/{{ .Values.image.name }}:{{ .Values.image.tag }}
+          imagePullPolicy: IfNotPresent
```
itaysk: isn't it better to remove this and let kubernetes put it's default (always for latest and ifnotpresent else)

deploy/kube-bench/values.yaml
```diff
@@ -0,0 +1,27 @@
...
+  name: kube-bench
+  tag: latest
+
+# runConfig is the configuration for running kube-bench
+runConfig:
+
+  # version is the Kubernetes version `--version` argument passed to kube-bench
+  # which is mandatory when run from inside a container since version
+  # auto-detection only works when kube-bench is run directly on the node.
+  version: 1.13
```
itaysk: this can be automated using `. Capabilities.KubeVersion`. it's ok if you prefer to leave it for another PR, just noting it here.

entrypoint.sh
```diff
@@ -15,5 +15,13 @@ if [ "$1" == "install" ]; then
     exit
   fi
 else
-  exec kube-bench "$@"
+  if [ -n "$SCHEDULE" ]; then
```
itaysk: would it make more sense to make schedule a flag rather than an environment variable? just from a UX perspective, this is an entirely different mode of operation (of the entrypoint)

entrypoint.sh
```diff
@@ -15,5 +15,13 @@ if [ "$1" == "install" ]; then
...
+    echo "$SCHEDULE" "cd $PWD && date && kube-bench" "$@" | crontab -c . -
+    crond -c . -f
+  else
+    kube-bench "$@"
+    # If the SCHEDULE variable is not set, the container exits.
+    # If set, the container sleeps.
+    [ -n "${SCHEDULE+set}" ] && while :; do sleep 86400; done
```
itaysk:
```suggestion
    [ -n "${SCHEDULE+set}" ] && while :; do sleep 1d; done
```

itaysk: what is the purpose of `+set` here?

deploy/kube-bench/templates/daemonset.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: kube-bench-ds
```
itaysk: wouldn't it be better to use the chart/release name here (`kube-bench.fullname` template)? (example in the generated templates when you run `helm create`)

deploy/kube-bench/values.yaml
s-nirali: I have not added this since the latest version supported by kube-bench could be different from the latest version of Kubernetes that a user could have installed.

entrypoint.sh
s-nirali: A CLI flag sounds like a good way. I could make this change if that's preferred.

entrypoint.sh
s-nirali: The Dockerfile had to accommodate two ways of running. 1) As a container in a pod as part of a Kubernetes deployment/daemonset 2) as a plain docker container that runs directly on the host. If `SCHEDULE` is set, it is assumed to be in mode 1 where the container sleeps after a single run if there is no schedule. This would prevent the daemonset from restarting the pod with an exited container over and over if the intention is to just run kube-bench once and then delete the deployment. If the variable is not set, it is assumed to run as a standalone container and it exits after running kube-bench

lizrice:
Hi @s-nirali, and thanks for everything you have done on this PR (and others). 

However, we've been talking about this issue internally and finally reached the conclusion that this isn't something we want to support at this time. We have reservations about the scheduling model, as it feels like it's having to work around the lack of native Kubernetes support for a DaemonSet / CronJob combination. We're also concerned about maintaining the helm chart and templating going forward - as they say, "no is temporary, but yes is forever" (and personally I have been caught out before with accepting what look like harmless functional additions, but turn out to add maintenance burden for years to come). 

If you wanted to support the helm chart elsewhere, we would be OK with referring to it from the README as a third-party addition. 

Sorry this is likely not the news you wanted to hear.

s-nirali:
Hello @lizrice . Thank you so much for reviewing and discussing this PR and I appreciate you giving me a feedback. I think I'll just let the PR itself contain this contribution and the [helm chart](https://github.com/s-nirali/kube-bench/tree/helm-chart/deploy/kube-bench) for whosoever needs something similar instead of maintaining it elsewhere, since this also needs a small update to the [Docker entrypoint](https://github.com/s-nirali/kube-bench/blob/128127a8690b049a5ea0152f6ea6d8d580156b59/entrypoint.sh) file.

PR #506, opened 2019-10-31
Updated Documentation A Lil Bit

vanshdevgan:
Added Last Line Which I Think Should Be There To Motivate Developers


simar7:
This is not needed. Thanks.

PR #514, opened 2019-11-06
docs: Note about not changing license etc

lizrice:


itaysk:
should we include issues/PR templates?

PR #516, opened 2019-11-07
Add option to output in JUnit format

### LINKED ISSUE #460: Output in junit
Would you be open to a PR which added the option to output in a junit format so that these could effectively be parsed as a test suite?

As of now the json output is just the json representation of the controls object which has to be parsed as a one-off if someone wants to automate this. If we could output into junit then other tools could more easily consume the results and check for pass/failure count etc.

johnSchnake:
If running these checks in a CI system it may be beneficial
to output in a more standardized format such as JUnit for
parsing by other tools in a consistent manner.

Fixes #460

Signed-off-by: John Schnake <jschnake@vmware.com>


johnSchnake:
I'm not sure what you're policy is on the code coverage but this is just a serialization routine so any test is effectively going to just be a rehash of the code. I can write a test confirming the output on a test run so it will just compare the serialization and you can see what it looks like if you want (e.g. a "goldenfile" test).

I submitted with no tests due to the lack of much real _logic_ here and since the other serialization routines didn't have tests of that nature either.

johnSchnake:
@DhruvTrehan You responded on the issue after I mentioned I put up a PR but it seemed like you still thought I was asking you to handle it.

This PR is ready to go; let me know what you think or what you'd need from me to get it over the finish line.

johnSchnake:
cc @robertojrojas I've seen that you are active on a number of tickets in the next milestone so I was hoping you could take a look. Hoping this could get into the milestone as well.

robertojrojas:
@johnSchnake thanks for the contribution.
 I'll review it and probably include it to the next release v0.2.0

johnSchnake:
Added tests and have a passing codecov patch now.

johnSchnake:
@robertojrojas Just for clarity, I see that the v0.2.0 milestone is set for today. Does that mean that you intend to push a release today? Or when do you think that release will be bumped to?

robertojrojas:
@johnSchnake We're working really hard to make the release today. 🤕

Issue #517, opened 2019-11-08
Besides Kubectl and Kubelet Use the REST API to Determines Kubernetes Version


robertojrojas:
Currently **kube-bench** depends on **kubectl** or **kubelet** as the only way to determine the version of Kubernetes.

I'm proposing that in addition to the above programs, kube-bench uses the **Kubernetes REST API** to determine the version of Kubernetes.

robertojrojas:
Fixed by #518 
Closing...

PR #518, opened 2019-11-08
Fixes issue #517: Determines Kubernetes version using the REST API

robertojrojas:
Fixes issue #517 

Determines Kubernetes version using the [REST API](https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/#without-using-a-proxy)

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -269,11 +269,10 @@ func getBenchmarkVersion(kubeVersion, benchmarkVersion string, v *viper.Viper) (
 
 	if isEmpty(benchmarkVersion) {
 		if isEmpty(kubeVersion) {
-			kv, err := getKubeVersion()
+			kubeVersion, err = getKubeVersion()
 			if err != nil {
 				return "", fmt.Errorf("Version check failed: %s\nAlternatively, you can specify the version with --version", err)
 			}
-			kubeVersion = kv
 		}
 
 		kubeToBenchmarkMap, err := loadVersionMapping(v)
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -235,17 +235,19 @@ func loadConfig(nodetype check.NodeType) string {
 }
 
 func mapToBenchmarkVersion(kubeToBenchmarkMap map[string]string, kv string) (string, error) {
+	kvOriginal := kv
 	cisVersion, found := kubeToBenchmarkMap[kv]
+	glog.V(2).Info(fmt.Sprintf("mapToBenchmarkVersion for k8sVersion: %q cisVersion: %q found: %t\n", kv, cisVersion, found))
 	for !found && (kv != defaultKubeVersion && !isEmpty(kv)) {
 		kv = decrementVersion(kv)
 		cisVersion, found = kubeToBenchmarkMap[kv]
-		glog.V(2).Info(fmt.Sprintf("mapToBenchmarkVersion for cisVersion: %q found: %t\n", cisVersion, found))
+		glog.V(2).Info(fmt.Sprintf("mapToBenchmarkVersion for k8sVersion: %q cisVersion: %q found: %t\n", kv, cisVersion, found))
 	}
 
 	if !found {
-		glog.V(1).Info(fmt.Sprintf("mapToBenchmarkVersion unable to find a match for: %q", kv))
+		glog.V(1).Info(fmt.Sprintf("mapToBenchmarkVersion unable to find a match for: %q", kvOriginal))
 		glog.V(3).Info(fmt.Sprintf("mapToBenchmarkVersion kubeToBenchmarkSMap: %#v", kubeToBenchmarkMap))
-		return "", fmt.Errorf("Unable to find a matching Benchmark Version match for kubernetes version: %s", kubeVersion)
+		return "", fmt.Errorf("unable to find a matching Benchmark Version match for kubernetes version: %s", kvOriginal)
 	}
 
 	return cisVersion, nil
@@ -267,10 +269,11 @@ func getBenchmarkVersion(kubeVersion, benchmarkVersion string, v *viper.Viper) (
 
 	if isEmpty(benchmarkVersion) {
 		if isEmpty(kubeVersion) {
-			kubeVersion, err = getKubeVersion()
+			kv, err := getKubeVersion()
 			if err != nil {
 				return "", fmt.Errorf("Version check failed: %s\nAlternatively, you can specify the version with --version", err)
 			}
+			kubeVersion = kv
 		}
 
 		kubeToBenchmarkMap, err := loadVersionMapping(v)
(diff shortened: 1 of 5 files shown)


cmd/common.go
```diff
@@ -235,17 +235,19 @@ func loadConfig(nodetype check.NodeType) string {
...
-		glog.V(2).Info(fmt.Sprintf("mapToBenchmarkVersion for cisVersion: %q found: %t\n", cisVersion, found))
+		glog.V(2).Info(fmt.Sprintf("mapToBenchmarkVersion for k8sVersion: %q cisVersion: %q found: %t\n", kv, cisVersion, found))
 	}
 
 	if !found {
-		glog.V(1).Info(fmt.Sprintf("mapToBenchmarkVersion unable to find a match for: %q", kv))
+		glog.V(1).Info(fmt.Sprintf("mapToBenchmarkVersion unable to find a match for: %q", kvOriginal))
 		glog.V(3).Info(fmt.Sprintf("mapToBenchmarkVersion kubeToBenchmarkSMap: %#v", kubeToBenchmarkMap))
-		return "", fmt.Errorf("Unable to find a matching Benchmark Version match for kubernetes version: %s", kubeVersion)
+		return "", fmt.Errorf("unable to find a matching Benchmark Version match for kubernetes version: %s", kvOriginal)
```
simar7: Is there a test for this returned error that needs to change?

cmd/kubernetes_version.go
```diff
@@ -0,0 +1,155 @@
...
+		return nil, err
+	}
+
+	authToken := fmt.Sprintf("Bearer %s", token)
+	glog.V(2).Info(fmt.Sprintf("getWebData AUTH TOKEN --[%q]--\n", authToken))
+	req.Header.Set("Authorization", authToken)
+
+	resp, err := client.Do(req)
+	if err != nil {
+		fmt.Printf("HTTP ERROR: %v\n", err)
```
simar7: Replace with the appropriate logger.

cmd/common.go
robertojrojas: Good catch! I'm thinking about reverting this change as it should be on its own issue and proper tests.

cmd/kubernetes_version.go
```diff
@@ -0,0 +1,155 @@
...
+
+	tlsConf := &tls.Config{
+		Certificates:       []tls.Certificate{*cacert},
+		InsecureSkipVerify: true,
+	}
+	tr := &http.Transport{
+		TLSClientConfig: tlsConf,
+	}
+	client := &http.Client{Transport: tr}
+	req, err := http.NewRequest("GET", srvURL, nil)
```
simar7: Instead of hardcoding `GET` use `http.MethodGet`.

cmd/kubernetes_version.go
robertojrojas: 👍

cmd/kubernetes_version.go
robertojrojas: 👍

cmd/kubernetes_version.go
```diff
@@ -0,0 +1,155 @@
...
+		  "minor": "15",
+		  "gitVersion": "v1.15.3",
+		  "gitCommit": "2d3c76f9091b6bec110a5e63777c332469e0cba2",
+		  "gitTreeState": "clean",
+		  "buildDate": "2019-08-20T18:57:36Z",
+		  "goVersion": "go1.12.9",
+		  "compiler": "gc",
+		  "platform": "linux/amd64"
+		}
+	*/
```
simar7: I think you already added a test for this to document what's proper JSON. We can remove this.

cmd/kubernetes_version.go
```diff
@@ -0,0 +1,155 @@
...
+	// The following provides flexibility to use
+	// K8S provided variables is situations where
+	// hostNetwork: true
+	if !isEmpty(os.Getenv("KUBE_BENCH_K8S_ENV")) {
+		k8sHost := os.Getenv("KUBERNETES_SERVICE_HOST")
+		k8sPort := os.Getenv("KUBERNETES_SERVICE_PORT_HTTPS")
+		if !isEmpty(k8sHost) && !isEmpty(k8sPort) {
+			k8sVersionURL = fmt.Sprintf("https://%s:%s/version", k8sHost, k8sPort)
+		} else {
+			glog.V(2).Info(fmt.Sprintf("KUBE_BENCH_K8S_ENV is set, but environment variables KUBERNETES_SERVICE_HOST or KUBERNETES_SERVICE_PORT_HTTPS are not set"))
```
simar7: What should we expect to happen if this branch was taken? Would https://github.com/aquasecurity/kube-bench/pull/518/files#diff-5c45b92badff1436a68538572a740210R33 ultimately return an error?

cmd/kubernetes_version.go
```diff
@@ -0,0 +1,155 @@
...
+	"strings"
+
+	"github.com/golang/glog"
+)
+
+func getKubeVersionFromRESTAPI() (string, error) {
+	k8sVersionURL := getKubernetesURL()
+	serviceaccount := "/var/run/secrets/kubernetes.io/serviceaccount"
+	cacertfile := fmt.Sprintf("%s/ca.crt", serviceaccount)
+	tokenfile := fmt.Sprintf("%s/token", serviceaccount)
```
simar7: Can we inject these as args to `getKubeVersionFromRESTAPI()` so we are able to test this method? Currently https://github.com/aquasecurity/kube-bench/pull/518/files#diff-5c45b92badff1436a68538572a740210R22-R42 are untested.

cmd/util.go
```diff
@@ -271,6 +275,12 @@ Alternatively, you can specify the version with --version
 `
 
 func getKubeVersion() (string, error) {
+
+	if k8sVer, err := getKubeVersionFromRESTAPI(); err == nil {
+		glog.V(2).Info(fmt.Sprintf("Kubernetes REST API Reported version: %s", k8sVer))
+		return k8sVer, nil
+	}
```
simar7: This looks untested as it stands. Should we add coverage? Might be better to look at this after we've addressed https://github.com/aquasecurity/kube-bench/pull/518/files#r344931996

cmd/kubernetes_version.go
robertojrojas: I didn't think I would need a test for **getKubeVersionFromRESTAPI()** as all the scenarios in it are already covered by the methods called .

cmd/common.go
robertojrojas: I modified that tests to check for the error string.

cmd/kubernetes_version.go
robertojrojas: if the environment variables are missing, it's fine. I would just try the default URL.

cmd/util.go
robertojrojas: The wrapping method that calls this method is difficult to test because there is a scenario that performs a search against the root (**'/'**) directory. 
I'm thinking about refactoring that function in the future, but I need to coordinate with Liz to minimize impact on the cilent.

simar7: Fair point.

cmd/kubernetes_version.go
simar7: You're right. After having a second look at this, I agree with you.

Issue #439, opened 2019-10-10
Expand Integration Testing Using Kind


robertojrojas:
The project already provides integration testing using `kind`, but there is some for improvements.
It'll be good to have a in-depth integration testing done with `kind`.

Initially, it'll provide ability to execute integration tests using **job.yaml**, **job-master.yaml**, and **job-node.yaml**.

Later on, more tests will be added.

robertojrojas:
Fixed by #520

PR #520, opened 2019-11-14
Fixes issue #439: Adds integration testing using KIND

robertojrojas:
Fixes issue #439: 

Adds integration testing using KIND


integration/integration.go
```diff
@@ -0,0 +1,184 @@
...
+					ticker := time.NewTicker(timer)
+					for {
+						fmt.Println("using ticker and an timer...")
+						select {
+						case <-ticker.C:
+							thePod, err := clientset.CoreV1().Pods(apiv1.NamespaceDefault).Get(cp.Name, metav1.GetOptions{})
+							if err != nil {
+								return nil, err
+							}
+							fmt.Printf("thePod (%s) - status:%#v reason:%s message:%s\n", thePod.Name, thePod.Status.Phase, thePod.Status.Reason, thePod.Status.Message)
```
simar7: The `reason` and `message` are always empty it seem when I ran it.

integration/integration.go
```diff
@@ -0,0 +1,184 @@
...
+						fmt.Printf("pod (%s) - %s - retrying...\n", cp.Name, cp.Status.Phase)
+						failedPods[cp.Name] = struct{}{}
+						break podfailed
+					}
+
+					// Pod still working
+					// Wait and try again...
+					ticker := time.NewTicker(timer)
+					for {
+						fmt.Println("using ticker and an timer...")
```
simar7: Suggestion: Maybe print this every `x` times rather than every tick? Currently the output is super verbose as a result.

integration/integration.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	return clientset, nil
+}
+
+func int32Ptr(i int32) *int32 { return &i }
```
simar7: This doesn't seem to be used. Remove?

integration/integration.go
```diff
@@ -0,0 +1,184 @@
...
+				}
+			}
+		}
+		time.Sleep(1 * time.Second)
+	}
+
+	return nil, fmt.Errorf("no Pod with %s", name)
+}
+
+func getPodLogs(clientset *kubernetes.Clientset, pod *apiv1.Pod) string {
```
simar7: What are your thoughts on returning (string, error) in case pod logs were not obtained? The caller of getPodLogs() can then decide appropriately what to do with it.

integration/integration.go
```diff
@@ -0,0 +1,184 @@
...
+	defer podLogs.Close()
+
+	buf := new(bytes.Buffer)
+	_, err = io.Copy(buf, podLogs)
+	if err != nil {
+		return "error in copy information from podLogs to buf"
+	}
+	str := buf.String()
+
+	return str
```
simar7: Could replace with `return buf.String()`

integration/integration.go
robertojrojas: Yeah I used to try to troubleshoot a pod in hanging mode. I'll remove it.

integration/integration.go
robertojrojas: I see your point, but I wanted getPodLogs to be a simple either got logs or string error. Since we're returning that string to the test method. If we decide to use this method for other purposes, we can refactor it to make it better.

integration/integration.go
robertojrojas: 👍

integration/integration.go
robertojrojas: It's ok for now. This method is already a bit complicated :)

integration/integration.go
robertojrojas: 👍

Issue #391, opened 2019-08-22
kube-bench version command does not work in docker image


ttousai:
kube-bench version: aquasec/kube-bench:latest
sha256: 719706c60ed5a08051c335d7654c65f4ab312957a2aec7805acbf4965794b556

Running the version command in the docker image for aquasec/kube-bench returns nothing rather than the version of kube-bench.

Reproduce:
```
$ sudo docker pull aquasec/kube-bench@sha256:719706c60ed5a08051c335d7654c65f4ab312957a2aec7805acbf4965794b556
$ sudo docker run aquasec/kube-bench:latest version
```

robertojrojas:
Fixed by #428

robertojrojas:
This issue is back or probably I didn't fix it correctly :(

robertojrojas:
Fixed by #522

PR #522, opened 2019-11-26
Fixes issue #391: Replaced calling docker directly by using "make build-docker"

robertojrojas:
Fixes issue #391: 

Replaced calling docker directly by using "make build-docker"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_install:
 
 script:
   - GO111MODULE=on go test ./...
-  - IMAGE_NAME=kube-bench KUBEBENCH_VERSION=$TRAVIS_BRANCH make build-docker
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -15,7 +15,7 @@ before_install:
 
 script:
   - GO111MODULE=on go test ./...
-  - docker build --tag kube-bench .
+  - IMAGE_NAME=kube-bench make build-docker
```
danielpacak: Is there any reason to keep IMAGE_NAME var outside of makefile?

robertojrojas: The use of the IMAGE_NAME is to make sure the line following that one continues to work.

danielpacak: All right, I thought that the defaults hardcoded in makefile are enough, so when I run `make build-docker` it would work the same way as on Travis.

PR #526, opened 2019-11-29
Openshift configs

caruccio:
Improves support for OKD

### DIFF
--- a/cfg/config.yaml
+++ b/cfg/config.yaml
@@ -25,6 +25,7 @@ master:
       - "hyperkube apiserver"
       - "hyperkube kube-apiserver"
       - "apiserver"
+      - "openshift start master api"
     confs:
       - /etc/kubernetes/manifests/kube-apiserver.yaml
       - /etc/kubernetes/manifests/kube-apiserver.manifest
@@ -37,6 +38,7 @@ master:
       - "hyperkube scheduler"
       - "hyperkube kube-scheduler"
       - "scheduler"
+      - "openshift start master controllers"
     confs:
       - /etc/kubernetes/manifests/kube-scheduler.yaml
       - /etc/kubernetes/manifests/kube-scheduler.manifest
@@ -50,6 +52,7 @@ master:
       - "hyperkube controller-manager"
       - "hyperkube kube-controller-manager"
       - "controller-manager"
+      - "openshift start master controllers"
     confs:
       - /etc/kubernetes/manifests/kube-controller-manager.yaml
       - /etc/kubernetes/manifests/kube-controller-manager.manifest
--- a/cfg/rh-0.7/config.yaml
+++ b/cfg/rh-0.7/config.yaml
@@ -22,6 +22,9 @@ master:
       - openshift start etcd
 
 node:
+  svcs:
+    - /etc/systemd/system/atomic-openshift-node.service
+    - /etc/systemd/system/origin-node.service
   proxy:
     bins:
       - openshift start network
--- a/cfg/rh-0.7/node.yaml
+++ b/cfg/rh-0.7/node.yaml
@@ -254,7 +254,7 @@ groups:
 
   - id: 8.3
     text: "Verify the kubelet service file permissions of 644"
-    audit: "stat -c %a /etc/systemd/system/atomic-openshift-node.service"
+    audit: "stat -c %a $nodesvc"
     tests:
       bin_op: or
       test_items:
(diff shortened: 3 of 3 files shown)


cfg/config.yaml
```diff
@@ -25,6 +25,7 @@ master:
       - "hyperkube apiserver"
       - "hyperkube kube-apiserver"
       - "apiserver"
+      - "openshift start master api"
```
robertojrojas: I was looking at OKD 3.11 and I noticed `hypershift openshift-kube-apiserver` instead of the above.

cfg/config.yaml
```diff
@@ -37,6 +38,7 @@ master:
       - "hyperkube scheduler"
       - "hyperkube kube-scheduler"
       - "scheduler"
+      - "openshift start master controllers"
```
robertojrojas: I was looking at OKD 3.11 and I noticed `hypershift openshift-controller-manager` instead of the above.

cfg/config.yaml
```diff
@@ -50,6 +52,7 @@ master:
       - "hyperkube controller-manager"
       - "hyperkube kube-controller-manager"
       - "controller-manager"
+      - "openshift start master controllers"
```
robertojrojas: I was looking at OKD 3.11 and I noticed `hypershift openshift-controller-manager` instead of the above.

caruccio: I believe this is the OCP (non-OKD) release. As you can see from links below for both 3.10 and 3.11:

api-server:

https://github.com/openshift/openshift-ansible/blob/release-3.10/roles/openshift_control_plane/files/apiserver.yaml#L26
https://github.com/openshift/openshift-ansible/blob/release-3.11/roles/openshift_control_plane/files/apiserver.yaml#L27

controllers:

https://github.com/openshift/openshift-ansible/blob/release-3.10/roles/openshift_control_plane/files/controller.yaml#L26
https://github.com/openshift/openshift-ansible/blob/release-3.11/roles/openshift_control_plane/files/controller.yaml#L27

caruccio: Or even an RPM based installation.

robertojrojas: I just tried this from within a Pod running on OKD 3.11.
kube-bench executes a PS commands using each of the values configured in the config.yaml file 
and parses the results to see if it finds a match.

If config.yaml is configured with `openshift start master api` kube-bench will execute the following command to find the API Server:
```
/opt/kube-bench # /bin/ps -C openshift  -o cmd --no-headers
openshift start etcd --config=/etc/origin/master/master-config.yaml
openshift start network --enable=proxy --listen=https://0.0.0.0:8444 --config=/etc/origin/node/node-config.yaml
openshift start network --enable=dns --config=/etc/origin/node/node-config.yaml
```

But, if `hypershift openshift-kube-apiserver` is used, kube-bench will execute the following command instead:
```
/opt/kube-bench # /bin/ps -C hypershift  -o cmd --no-headers
hypershift openshift-kube-apiserver --config=/etc/origin/master/master-config.yaml
hypershift openshift-controller-manager --config=/etc/origin/master/master-config.yaml --v=0
hypershift openshift-apiserver --config=/etc/origin/master/master-config.yaml -v=0
hypershift experimental openshift-webconsole-operator --config=/var/run/configmaps/config/operator-config.yaml -v=0
```
As you can see in the output using `hypershift openshift-kube-apiserver`, kube-bench will have a better chance
matching the API Server program.

caruccio: My point is that there are more than one okd installer versions out there: the open source one (upstream, deployed by github.com/openshift/openshift-ansible) and the managed/enterprise one (which I've no access to) from redhat subscription (I guess).

In fact I recall seen `hypershift` from okd <3.9.

I've plenty of okd clusters, both 3.10 and 3.11, wthich gives the same results:

```
$ oc version
oc v3.10.0+0c4577e-1
kubernetes v1.10.0+b81c8f8
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://XXXXXXX:443
openshift v3.10.0+6c78c44-115
kubernetes v1.10.0+b81c8f8

$ /bin/ps -C hypershift  -o cmd --no-headers|wc -l
0

$ /bin/ps -C openshift  -o cmd --no-headers
openshift start master controllers --config=/etc/origin/master/master-config.yaml --listen=https://0.0.0.0:8444 --loglevel=3
openshift start master api --config=/etc/origin/master/master-config.yaml --loglevel=3
openshift start network --config=/etc/origin/node/node-config.yaml --kubeconfig=/tmp/kubeconfig --loglevel=1

```

There isn't any reference to `hypershift` from the sources:

```
$ git remote -v
origin  https://github.com/openshift/openshift-ansible.git (fetch)
origin  https://github.com/openshift/openshift-ansible.git (push)
$ git branch
* (HEAD detached at dbc256b74)
  master
$ grep hypershift * -rw | wc -l
0
```

What if I simply add `hypershift` to configs? That way we can catch both installation methods.
Does it sounds good enough?

robertojrojas: I stand corrected! You're right about the use of `openshift start master ...` 
I just verified that this is how the control plane processes appear in OCP 3.11.
Now,  as to OKD 3.11, I think the `hypershift openshift-kube-apiserver`, `hypershift openshift-controller-manager`,  and `hyperkube kube-scheduler` should be add as well.

Since these values are specific to OCP, the must be add in **cfg/rh-0.7/config.yaml** instead of **cfg/config.yaml**

Issue #494, opened 2019-10-25
support for CIS Benchmark version 1.5 


winnettjohnston:
Great tool! The kubernetes CIS benchmark 1.5 was released. What is the timeline to align with the new benchmark?

jcmcken:
Also interested in this.

robertojrojas:
**_Note_** that we changed the **cfg** directory layout to use the CIS Benchmark version instead of the Kubernetes as it used to be.
So, the CIS 1.5 Benchmark will be in the **`cfg/cis-1.5`** directory. The configuration files inside and their names will stay the same as before - _config.yaml, node.yaml, master.yaml_

leodotcloud:
@robertojrojas with 1.5, etcd has been pulled out under it's own section. I guess we would need _etcd.yaml_ along with the existing _master.yaml_ and _node.yaml_.

Also, we have master and node as sub commands to the top level CLI (kube-bench). With the introduction of etcd section, we might need a new sub command added as well. Or another option would be to pass the role as an argument instead of a sub command and use the name of the role to fetch the corresponding yaml file. With this approach, let's say in future we have a new section (ingress for example) we can easily extend the CLI to accommodate it.

robertojrojas:
@leodotcloud We're still looking into CIS 1.5. I'm not completely sure that's the approach the tool will follow. We'll see. :)

leodotcloud:
Cool. BTW, one more pointer/suggestion :) 

We can reuse the master sub command and pass the controls file using `-f`. And then have the etcd related controls in it's own file to run the following:

For master: `kube-bench master -f master.yaml --outputfile master.json`
For etcd: `kube-bench master -f etcd.yaml --outputfile etcd.json`

lizrice:
I think I have a proposal that is back-compatible with today's behaviour. 

CIS 1.5 has the following sections: 

1. Master (well, this is actually called Control Plane Components but this is too close to section 3) 
2. Etcd
3. Control Plane
4. Worker 
5. Policies

We could refer to these with the following names: `master`, `etcd`, `controlplane`, `worker`, `policies` and have the test definitions in corresponding YAML `master.yaml`, `etcd.yaml` etc. 

Extend the `kube-bench` sub-commands so that in addition to `master` and `node` we accept all these names (where `worker` and `node` are equivalent).  So, 
- `kube-bench etcd` runs the tests from `etcd.yaml`
- `kube-bench master` runs the tests from `master.yaml`
etc

[...]

PR #530, opened 2019-12-03
Fixes Issue #494 - add tests for CIS 1.5

robertojrojas:
Fixes Issue #494
Thanks @ttousai for getting most of the work done on this. 👍


cfg/cis-1.5/etcd.yaml
```diff
@@ -0,0 +1,131 @@
+---
+controls:
+version: 1.15
+id: 2
+text: "etcd"
+type: "master"
+groups:
+- id: 2
+  text: ""
```
lizrice: Should there be some text here?

cfg/cis-1.5/etcd.yaml
```diff
@@ -0,0 +1,131 @@
+---
+controls:
+version: 1.15
+id: 2
+text: "etcd"
+type: "master"
```
lizrice: Is this the right type? Etcd doesn’t necessarily run on the same node as the other control plane (master) components, which is why these tests got split out to a separate section in the CIS benchmark

cmd/run.go
```diff
@@ -66,20 +70,31 @@ func run(sections []string, benchmarkVersion string) (err error) {
 	return nil
 }
 
-func getTestYamlFiles(sections []string, benchmarkVersion string) (yamlFiles []string, err error) {
+func providesTargets(benchmarkVersion string) bool {
```
lizrice: Not sure about this function name, looks like it is getting the benchmark version number rather than checking targets? 

I was going to say that I don’t see where this is checking the targets, but I suspect you are thinking that there aren’t targets for CIS < 1.5.

All versions do provide some targets - before 1.5 the targets were master and node, and now there are some additional ones. (Otherwise you’d only be able to use the —targets option with >=1.5 which seems an unnecessary restriction)

cfg/cis-1.5/etcd.yaml
robertojrojas: That would be my guess too. I'll see if I can find something suitable to add there.

robertojrojas:
> Mostly looks OK though I do have a few comments.
> 
> This doesn’t have the auto-detection part, right?
I added auto-detection with conditions so it runs the correct `targets` based on the CIS benchmark version.

README.md
```diff
@@ -42,12 +42,13 @@ Table of Contents
...
 
-kube-bench supports the tests for Kubernetes as defined in the CIS Benchmarks 1.3.0 to 1.4.1 respectively. 
+kube-bench supports the tests for Kubernetes as defined in the CIS Benchmarks 1.3.0 to 1.5 respectively. 
 
 | CIS Kubernetes Benchmark | kube-bench config | Kubernetes versions |
 |---|---|---|
 | 1.3.0| cis-1.3 | 1.11-1.12 |
-| 1.4.1| cis-1.4 | 1.13- |
+| 1.4.1| cis-1.4 | 1.13-1.14 |
+| 1.5| cis-1.5 | 1.15- |
```
lizrice:
```suggestion
| 1.5.0 | cis-1.5 | 1.15- |
```

README.md
```diff
@@ -42,12 +42,13 @@ Table of Contents
       
 ## CIS Kubernetes Benchmark support
 
-kube-bench supports the tests for Kubernetes as defined in the CIS Benchmarks 1.3.0 to 1.4.1 respectively. 
+kube-bench supports the tests for Kubernetes as defined in the CIS Benchmarks 1.3.0 to 1.5 respectively.
```
lizrice:
```suggestion
kube-bench supports the tests for Kubernetes as defined in the CIS Benchmarks 1.3.0 to 1.5.0 respectively. 
```

Might as well be consistent in using the full version number

README.md
```diff
@@ -97,6 +98,28 @@ Alternatively, you can specify `--benchmark` to run a specific CIS Benchmark ver
...
+The following table shows the valid targets based on the CIS Benchmark version.
+| CIS Benchmark | Targets |
+|---|---|
+| cis-1.3| master, node |
+| cis-1.4| master, node |
+| cis-1.5| master, controlplane, node, etcd, policies |
+
+If no targets are specified, `kube-bench` will determine the appropriate targets based on the CIS Benchmark version.
+
+**Note:**  **`It is an error to specify invalid targets `**
```
lizrice: I'm not sure this is worth saying - it's not wrong, but we don't spell out all the other possible things that would be invalid input. (I feel differently about --version and --benchmark because unless we say something, it's not obvious whether we would treat both being supplied as an error - we might have decided differently)

PR #533, opened 2019-12-06
docs: Fix table rendering in README

lizrice:


lizrice:
Looks like this already got fixed elsewhere

Issue #535, opened 2019-12-06
Auto-detect Fails on OpenShift


robertojrojas:
When running **kube-bench** in the following manner:
```
# kube-bench --version ocp-3.11
```

it fails with the following error:
```console
exit status 127
```

robertojrojas:
Fixed by #537

PR #537, opened 2019-12-09
Fixes Issue #535

robertojrojas:
Fixes Issue #535

- Changes isEtcd condition so it only executes for CIS tests and not OpenShift tests.

- Adds `openssl`

- Adds programs for OpenShift OKD when installed using Minishift.

### DIFF
--- a/cfg/config.yaml
+++ b/cfg/config.yaml
@@ -25,7 +25,6 @@ master:
       - "hyperkube apiserver"
       - "hyperkube kube-apiserver"
       - "apiserver"
-      - "openshift start master api"
     confs:
       - /etc/kubernetes/manifests/kube-apiserver.yaml
       - /etc/kubernetes/manifests/kube-apiserver.manifest
@@ -38,7 +37,6 @@ master:
       - "hyperkube scheduler"
       - "hyperkube kube-scheduler"
       - "scheduler"
-      - "openshift start master controllers"
     confs:
       - /etc/kubernetes/manifests/kube-scheduler.yaml
       - /etc/kubernetes/manifests/kube-scheduler.manifest
@@ -52,7 +50,6 @@ master:
       - "hyperkube controller-manager"
       - "hyperkube kube-controller-manager"
       - "controller-manager"
-      - "openshift start master controllers"
     confs:
       - /etc/kubernetes/manifests/kube-controller-manager.yaml
       - /etc/kubernetes/manifests/kube-controller-manager.manifest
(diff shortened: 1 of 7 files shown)


cfg/config.yaml
```diff
@@ -25,7 +25,6 @@ master:
       - "hyperkube apiserver"
       - "hyperkube kube-apiserver"
       - "apiserver"
-      - "openshift start master api"
```
simar7: For my own learning: Why are these removed?

cmd/root.go
```diff
@@ -81,7 +81,7 @@ var RootCmd = &cobra.Command{
 
 		// Etcd is only valid for CIS 1.5 and later,
 		// this a gatekeeper for previous versions.
-		if isEtcd() && validTargets(benchmarkVersion, []string{string(check.ETCD)}) {
+		if validTargets(benchmarkVersion, []string{string(check.ETCD)}) && isEtcd() {
```
simar7: Does the order matter?

cmd/common_test.go
```diff
@@ -410,11 +491,6 @@ func fakeExecutableInPath(execFile, execCode string) (restoreFn, error) {
 		return nil, err
 	}
 
-	err = os.Chdir(tmp)
```
simar7: Curious why this was removed?

cfg/config.yaml
robertojrojas: Basically, these values are specific to the OpenShift, so they belong in the **cfg/rh-0.7/config.yaml** instead.

cmd/root.go
robertojrojas: Yes,  CIS 1.4 and earlier do not include `etcd.yaml`, so checking to see if it's provided by the CIS benchmark before trying to check if the host is running **Etcd** avoids misleading errors.

cmd/common_test.go
robertojrojas: It turns out I added the by mistake. We didn't have to navigate to the newly created `tmp` directory for the logic to work as the code refers to it using absolute paths. Removing it avoids errors with other Unit Tests that rely on being executed at the root of the source tree to find directories like `cfg`, etc...

cfg/rh-0.7/config.yaml
```diff
@@ -10,12 +10,14 @@ master:
   scheduler:
     bins:
       - "openshift start master controllers"
+      - "hyperkube kube-scheduler"
```
mrueg: @robertojrojas this should probably be hybershift instead of hyperkube, shouldn't it?

Issue #538, opened 2019-12-09
Add More Information on `exitWithError`


robertojrojas:
There are situation where the code only contains **`exitWithError(err)`**.  
When kube-bench is executed and one of those statements is executed, the user only gets an error without any further information.

robertojrojas:
Fixed by #539

PR #539, opened 2019-12-09
Fixes Issue #538

robertojrojas:
Fixes Issue #538

### DIFF
--- a/check/check.go
+++ b/check/check.go
@@ -180,7 +180,7 @@ func (c *Check) run() State {
 	}
 
 	if finalOutput != nil {
-		glog.V(3).Infof("Check.ID: %s Command: %q TestResult: %t Score: %q \n", c.ID, lastCommand, finalOutput.testResult, c.State)
+		glog.V(3).Infof("Check.ID: %s Command: %q TestResult: %t State: %q \n", c.ID, lastCommand, finalOutput.testResult, c.State)
 	} else {
 		glog.V(3).Infof("Check.ID: %s Command: %q TestResult: <<EMPTY>> \n", c.ID, lastCommand)
 	}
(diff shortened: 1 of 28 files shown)


check/check.go
```diff
@@ -180,7 +179,9 @@ func (c *Check) run() State {
 		}
 	}
 
-	if finalOutput == nil {
+	if finalOutput != nil {
+		glog.V(3).Infof("Check.ID: %s Command: %q TestResult: %t Score: %q \n", c.ID, lastCommand, finalOutput.testResult, c.State)
```
danielpacak: I see that you want to print score but you pass `c.State` as arg?

check/check.go
```diff
@@ -242,7 +243,7 @@ func isShellCommand(s string) bool {
 
 	out, err := cmd.Output()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "%s\n", err)
+		fmt.Fprintf(os.Stderr, "failed to check if command: %q is valid %v\n", s, err)
 		os.Exit(1)
```
danielpacak: NIT: To facilitate testing I'd return a wrapped error rather than call os.Exit here.

cmd/root.go
```diff
@@ -64,7 +64,7 @@ var RootCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		benchmarkVersion, err := getBenchmarkVersion(kubeVersion, benchmarkVersion, viper.GetViper())
 		if err != nil {
-			exitWithError(err)
+			exitWithError(fmt.Errorf("unable to get benchmark version. error: %v", err))
```
danielpacak: NIT: I don't like that we os.Exit in such common code / package. At some point it should be refactored to return wrapped errors so you could properly test it. I know it's not easy though.

For now you could have `exitWithErrorf("unable to get benchmark version: %v", err)`

check/check.go
robertojrojas: Yes, that's a typo. Good catch!

cmd/root.go
robertojrojas: Yeah, we're using that function a lot in **kube-bench**. A bit of refactoring would need to be done. For this PR, I just wanted to add a bit more information to help user troubleshooting.

Issue #536, opened 2019-12-09
RPM only contains the binary


rrey:
Hello,

The RPM only contains the `kube-bench` binary.
The configuration and yaml representation of the cis bench is not included and requires additional downloads to be able to run `kube-bench`.

Is there any reason for omitting this ?

robertojrojas:
@rrey thanks for reporting this. 
I'll take a look at cause of this and fix it.

robertojrojas:
Fixed by #540

PR #540, opened 2019-12-10
Fixes issue #536

robertojrojas:
Fixes issue #536

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,5 +1,6 @@
 env:
   - GO111MODULE=on
+  - KUBEBENCH_CFG=/etc/kube-bench/cfg
 builds:
   - main: main.go
     binary: kube-bench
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -1,20 +1,28 @@
...
   format: tar.gz
+  files:
+    - "cfg/**/*"
 nfpm:
   vendor: Aqua Security
   description: "The Kubernetes Bench for Security is a Go application that checks whether Kubernetes is deployed according to security best practices"
   license: Apache-2.0
   homepage: https://github.com/aquasecurity/kube-bench
+  files:
+    "cfg/**/*": "/etc/kube-bench/cfg"
```
rrey: I wonder if `cfg` is still required if files are already in `/etc/kube-bench` ?

You may want to align the location of the configuration between container and rpm (and other). I can see in the README that the container expects the files in `/opt/kube-bench/cfg/`.

IMHO: `/etc/kube-bench` sounds a better place to me.

Issue #543, opened 2019-12-21
Pod Security Policy tests are incorrectly marked as scored: false in cis-1.4


sowmyav27:
The following PodSecurityPolicies tests are marked as scored: false in kube-bench v cis-1.4
```
1.7.2
1.7.3
1.7.4
1.7.5
```
https://github.com/aquasecurity/kube-bench/blob/master/cfg/cis-1.4/master.yaml

```
- id: 1.7.2
    text: "Do not admit containers wishing to share the host process ID namespace (Not Scored)"
    type: "manual"
    remediation: |
     [Manual test]
     Create a PSP as described in the Kubernetes documentation, ensuring that the .spec.hostPID field is omitted or set to false.
    scored: false

  - id: 1.7.3
    text: "Do not admit containers wishing to share the host IPC namespace (Not Scored)"
    type: "manual"
    remediation: |
      [Manual test]
      Create a PSP as described in the Kubernetes documentation, ensuring that the .spec.hostIPC field is omitted or set to false.
    scored: false

  - id: 1.7.4
    text: "Do not admit containers wishing to share the host network namespace (Not Scored)"
    type: "manual"
    remediation: |
      [Manual test]
      Create a PSP as described in the Kubernetes documentation, ensuring that the .spec.hostNetwork field is omitted or set to false.
    scored: false

  - id: 1.7.5
    text: " Do not admit containers with allowPrivilegeEscalation (Not Scored)"
    type: "manual"
    remediation: |
      [Manual test]
      Create a PSP as described in the Kubernetes documentation, ensuring that the .spec.allowPrivilegeEscalation field is omitted or set to false.
    scored: false
```
The same tests are marked as "Scored" in the CIX Kubernetes Benchmark v1.4

**Expected behavior:**
The above mentioned tests be marked scored: true

Issue #542, opened 2019-12-20
Mismatch between the tests on kube-bench vs CIS Kubernetes Benchmark v1.4


sowmyav27:
**as per Kube-bench -**  https://github.com/aquasecurity/kube-bench/blob/master/cfg/cis-1.4/master.yaml
```
- id: 1.1.30
    text: "Ensure that the API Server only makes use of Strong Cryptographic Ciphers (Not Scored)"

  - id: 1.1.31
    text: "Ensure that the --etcd-cafile argument is set as appropriate (Scored)"

```

**As per CIS Kubernetes Benchmark v1.4**

```
1.1.30 Ensure that the --etcd-cafile argument is set as appropriate (Scored)
1.1.31 Ensure that the API Server only makes use of Strong Cryptographic Ciphers (Not Scored)
```

**Expected behavior:**
kube-bench should comply with CIS Kubernetes Benchmark v1.4

PR #544, opened 2019-12-23
fix mismatching checks

leodotcloud:
addresses:
- https://github.com/aquasecurity/kube-bench/issues/542
- https://github.com/aquasecurity/kube-bench/issues/543

### DIFF
--- a/cfg/cis-1.3/master.yaml
+++ b/cfg/cis-1.3/master.yaml
@@ -1437,7 +1437,7 @@ groups:
         scored: false
 
       - id: 1.7.2
-        text: "Do not admit containers wishing to share the host process ID namespace (Not Scored)"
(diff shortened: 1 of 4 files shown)


cfg/cis-1.3/master.yaml
```diff
@@ -493,6 +493,21 @@ groups:
     scored: true
 
   - id: 1.1.30
```
lizrice: In the CIS Kubernetes 1.3 benchmark they were the other way around i.e. 1.1.30 was about API Server, and 1.1.31 was about --etcd-cafile, so the change shouldn't be applied in this 1.3 version of the file.  (I guess this explains why our tests are the wrong way around in 1.4, though I don't know why they got swapped over in the CIS benchmark!)

leodotcloud:
@lizrice thank for catching the 1.3 changes! Fixed, could you plz re-review?

Issue #546, opened 2019-12-23
logtostderr is set to true always


leodotcloud:
```
kube-bench --help | grep logtostderr
      --logtostderr                      log to standard error instead of files (default false)
```

https://github.com/aquasecurity/kube-bench/blob/master/cmd/root.go#L105

PR #548, opened 2019-12-24
remove always true for logtostderr

leodotcloud:
addresses https://github.com/aquasecurity/kube-bench/issues/546


leodotcloud:
Thank you @lizrice for the review, updated README as requested.

Issue #552, opened 2020-01-03
Integration tests: re-use the same Kind control plane for all tests


lizrice:
in `TestRunWithKind` there are three different tests which all use the same Kind configuration. It's time consuming for the control plane to come up, so why not just set it up once and re-use it for all three tests? 

Maybe you were thinking of future-proofing for the possibility of having several different Kind configs. There could be an outer loop which works through the Kind configs (of which there is currently only one) and an inner loop which runs the different job tests.

PR #555, opened 2020-01-06
Integration Test: Improves performance and Reliability

### LINKED ISSUE #551: Integration tests: why do pods fail sometimes?

```
using ticker and an timer...
using ticker and an timer...
using ticker and an timer...
Found (2) pods
using ticker and an timer...
```

robertojrojas:
Fixes issues #552
 Improves performance and reliability.

### DIFF
--- a/integration/integration.go
+++ b/integration/integration.go
@@ -19,16 +19,7 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster/create"
 )
 
-func runWithKind(clusterName, kindCfg, kubebenchYAML, kubebenchImg string, timeout, ticker time.Duration) (string, error) {
-	options := create.WithConfigFile(kindCfg)
-	ctx := cluster.NewContext(clusterName)
-	if err := ctx.Create(options); err != nil {
-		return "", err
-	}
-	defer func() {
-		ctx.Delete()
-	}()
-
+func runWithKind(ctx *cluster.Context, jobName, kubebenchYAML, kubebenchImg string, timeout, ticker time.Duration) (string, error) {
 	clientset, err := getClientSet(ctx.KubeConfigPath())
 	if err != nil {
 		return "", err
@@ -50,27 +41,36 @@ func runWithKind(clusterName, kindCfg, kubebenchYAML, kubebenchImg string, timeo
 	}
 	job.Spec.Template.Spec.Containers[0].Image = kubebenchImg
 
-	if err := loadImageFromDocker(kubebenchImg, ctx); err != nil {
-		return "", err
-	}
-
 	_, err = clientset.BatchV1().Jobs(apiv1.NamespaceDefault).Create(job)
 	if err != nil {
 		return "", err
 	}
 
-	clientset, err = getClientSet(ctx.KubeConfigPath())
-	if err != nil {
-		return "", err
-	}
-
-	p, err := findPodForJob(clientset, "kube-bench", timeout, ticker)
+	p, err := findPodForJob(clientset, jobName, timeout, ticker)
 	if err != nil {
 		return "", err
 	}
 
 	output := getPodLogs(clientset, p)
+
+	err = clientset.BatchV1().Jobs(apiv1.NamespaceDefault).Delete(jobName, nil)
+	if err != nil {
+		return "", err
+	}
+
 	return output, nil
+}
+
+func setupCluster(clusterName, kindCfg string, duration time.Duration) (*cluster.Context, error) {
+	options := create.WithConfigFile(kindCfg)
+	toptions := create.WaitForReady(duration)
+	ctx := cluster.NewContext(clusterName)
+	if err := ctx.Create(options, toptions); err != nil {
+		return nil, err
+	}
+
+	return ctx, nil
+
 }
 
 func getClientSet(configPath string) (*kubernetes.Clientset, error) {
@@ -117,41 +117,6 @@ func findPodForJob(clientset *kubernetes.Clientset, name string, tout, timer tim
 						break podfailed
 					}
 
-					// Pod still working
-					// Wait and try again...
-					ticker := time.NewTicker(timer)
-					for {
-						fmt.Println("using ticker and an timer...")
-						select {
-						case <-ticker.C:
-							thePod, err := clientset.CoreV1().Pods(apiv1.NamespaceDefault).Get(cp.Name, metav1.GetOptions{})
-							if err != nil {
-								return nil, err
-							}
-							fmt.Printf("thePod (%s) - status:%#v \n", thePod.Name, thePod.Status.Phase)
-							if thePod.Status.Phase == apiv1.PodSucceeded {
-								return thePod, nil
-							}
-
-							if thePod.Status.Phase == apiv1.PodFailed {
-								fmt.Printf("thePod (%s) - %s - retrying...\n", thePod.Name, thePod.Status.Phase)
-								failedPods[thePod.Name] = struct{}{}
-								ticker.Stop()
-								break podfailed
-							}
-
-							if thePod.Status.Phase == apiv1.PodPending && strings.Contains(thePod.Status.Reason, "Failed") {
-								fmt.Printf("thePod (%s) - %s - retrying...\n", thePod.Name, thePod.Status.Reason)
-								failedPods[thePod.Name] = struct{}{}
-								ticker.Stop()
-								break podfailed
-							}
-
-						case <-timeout:
-							ticker.Stop()
-							return nil, fmt.Errorf("getPod time out: no Pod with %s", name)
-						}
-					}
 				}
 			}
 		}
--- a/integration/integration.go
+++ b/integration/integration.go
@@ -19,29 +19,8 @@ import (
 	"sigs.k8s.io/kind/pkg/cluster/create"
 )
 
-func runWithKind(ctx *cluster.Context, jobName, kubebenchYAML, kubebenchImg string, timeout time.Duration) (string, error) {
-	clientset, err := getClientSet(ctx.KubeConfigPath())
-	if err != nil {
-		return "", err
-	}
-
-	jobYAML, err := ioutil.ReadFile(kubebenchYAML)
-	if err != nil {
-		return "", err
-	}
-
-	decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(jobYAML), len(jobYAML))
-	if err != nil {
-		return "", err
-	}
-
-	job := &batchv1.Job{}
-	if err := decoder.Decode(job); err != nil {
-		return "", err
-	}
-	job.Spec.Template.Spec.Containers[0].Image = kubebenchImg
-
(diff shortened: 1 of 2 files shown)


integration/integration.go
```diff
@@ -117,41 +117,6 @@ func findPodForJob(clientset *kubernetes.Clientset, name string, tout, timer tim
 						break podfailed
 					}
 
-					// Pod still working
```
lizrice: (GitHub won't let me put comments in code that hasn't changed, so sorry this isn't right next to the code I'm referring to.)

I don't think it's worth spending the time to change now, but there are two things I don't love about this:
- It's querying across all the pods in the namespace - why not just look at the job? 
- As a general rule, polling has a bad smell and it's much better to be event-driven. I think you probably needed [`JobInformer`](https://github.com/kubernetes/client-go/blob/master/informers/batch/v1/job.go)

lizrice: Lines 114-118 cover the case of the pod failing, so I don't think you have addressed #551 yet. Do you expect these jobs (and the pods within them) to ever fail except in cases where the whole integration test should fail?  

I would have thought that if a job fails, it should just fail the integration test. Or is there some condition under which we expect kube-bench to fail temporarily? If there is, at a minimum we should comment it and really we should be checking that the failure only happens in that condition.

integration/integration.go
```diff
@@ -117,41 +117,6 @@ func findPodForJob(clientset *kubernetes.Clientset, name string, tout, timer tim
...
-								ticker.Stop()
-								break podfailed
-							}
-
-						case <-timeout:
-							ticker.Stop()
-							return nil, fmt.Errorf("getPod time out: no Pod with %s", name)
-						}
-					}
 				}
```
lizrice: I am happy to see this second loop go away!

integration/integration_test.go
```diff
@@ -39,32 +38,41 @@ func TestRunWithKind(t *testing.T) {
...
+	}()
+
+	if err := loadImageFromDocker(*kubebenchImg, ctx); err != nil {
+		t.Fatalf("failed to load kube-bench image from Docker to KIND error: %v", err)
+	}
+
 	for _, c := range cases {
-		t.Run(c.TestName, func(t *testing.T) {
-			data, err := runWithKind(c.TestName, c.KindCfg, c.KubebenchYAML, *kubebenchImg, timeout, ticker)
+		t.Run(c.TestName, func(tc *testing.T) {
```
lizrice: I think it's more conventional to continue to call it `t` rather than `tc` inside the test function (as you did before)

lizrice:
Also let's have meaningful names for PRs, not just references to issue numbers

integration/integration_test.go
```diff
@@ -12,22 +12,21 @@ import (
...
+var timeout = flag.Duration("timeout", 10*time.Minute, "Test Timeout")
 
 func TestRunWithKind(t *testing.T) {
 	flag.Parse()
 	fmt.Printf("kube-bench Container Image: %s\n", *kubebenchImg)
-	timeout := time.Duration(10 * time.Minute)
-	ticker := time.Duration(2 * time.Second)
 
-	mustMatch := func(expFname, data string) {
+	mustMatch := func(tc *testing.T, expFname, data string) {
```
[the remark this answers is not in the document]
robertojrojas: I needed to pass the `tc` down to the function because using the enclosed variable `t` was causing the test to send a false positive for the current test like:
```
pod (kube-bench-zx2nk) - "Succeeded"
--- FAIL: TestRunWithKind (109.67s)
    integration_test.go:29: expected:
    ....
    ....
    --- PASS: TestRunWithKind/kube-bench (15.23s)
FAIL
FAIL    github.com/aquasecurity/kube-bench/integration  109.773s
FAIL
make: *** [integration-tests] Error 1
```

instead of 


```
-- FAIL: TestRunWithKind (87.64s)
    --- FAIL: TestRunWithKind/kube-bench (8.21s)
        integration_test.go:29: expected: 
        ...
        ...
FAIL
FAIL    github.com/aquasecurity/kube-bench/integration  87.695s
FAIL
make: *** [integration-tests] Error 1
```

integration/integration_test.go
```diff
@@ -39,32 +38,41 @@ func TestRunWithKind(t *testing.T) {
...
-			data, err := runWithKind(c.TestName, c.KindCfg, c.KubebenchYAML, *kubebenchImg, timeout, ticker)
+		t.Run(c.TestName, func(tc *testing.T) {
+			data, err := runWithKind(ctx, c.TestName, c.KubebenchYAML, *kubebenchImg, *timeout)
 			if err != nil {
-				t.Fatalf("unexpected error: %v", err)
-				return
+				tc.Errorf("unexpected error: %v", err)
 			}
-			mustMatch(c.ExpectedFile, data)
+			mustMatch(tc, c.ExpectedFile, data)
```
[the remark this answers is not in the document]
robertojrojas: As I explained above.

integration/integration.go
robertojrojas: It turns out that the pods were failing because the control plane was not completely ready by the time the job is deployed. `kube-bench` was failing trying to find a way to get the `Kubernetes` version. This is because the API Server was not accepting requests and then tried to use `kubectl`, which is a symlink in `KIND` and failed to execute.  To fix #551,  I used a function provided by `KIND` to wait until the open control plane is "ready". One thing I noticed is that in most situations when `KIND` reports that the control plane is ready, the first pod deployed works, but sometimes, the first pod fails and the second pod success. This is why I have loop expecting failed pods and trying again. I'll give the [JobInformer](https://github.com/kubernetes/client-go/blob/master/informers/batch/v1/job.go) a try.

integration/integration_test.go
robertojrojas: I'm using `tc` (as in test case) so that it's clear I'm not using the outer `t`.

integration/integration.go
robertojrojas: I looked at the **`JobInformer`** and I believe using it would add complexity without providing much more value than just performing a poll a few times per job to find the corresponding.

1. As far as I can tell the **`JobInformer`** provides [Job](https://github.com/kubernetes/api/blob/7dc09db16fb8/batch/v1/types.go#L28) objects without providing the name of the actual pod that successfully executed **`kube-bench`**, and then a search for the pod would need to be performed.
2. In the case of performance improvement, the bulk of the time is taken by `KIND` brining the Kubernetes cluster up, about 2 minutes.
3. Finding the Pods the first time takes around 8-10 seconds, but finding the pods for subsequent jobs only takes a fraction of that.
4. I added a Label Selector to find the Pod so the Kubernetes API query would be job specific.

lizrice: I agree, as I said in my comment I don't think it's worth spending time to change to JobInformer at this point, just that it would have been a cleaner way to do it

integration/integration_test.go
lizrice: The convention is to use `tc` for the test case (see https://blog.golang.org/subtests for an example), and to use `t` for the testing structure. These conventions make code easier to read and re-use, so let's stick with them

integration/integration_test.go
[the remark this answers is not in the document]
lizrice: The root of this problem is that `mustMatch()` is outside the enclosure for the version of `t` that's actually used in the tests

integration/integration_test.go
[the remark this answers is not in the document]
lizrice: Move `mustMatch()` into the right enclosure

integration/integration.go
lizrice: > It turns out that the pods were failing because the control plane was not completely ready by the time the job is deployed. kube-bench was failing trying to find a way to get the Kubernetes version. This is because the API Server was not accepting requests...

I don't understand this - you use the API from here to create the Job that runs the test, which starts a pod which starts kube-bench. So kube-bench can't be attempting API requests until some time after this test code has successfully made requests of the API Server, right?

integration/integration_test.go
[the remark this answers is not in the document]
robertojrojas: will do.

integration/integration_test.go
robertojrojas: sounds good.

integration/integration.go
robertojrojas: It's odd that from time to time, even after waiting for the `KIND` to indicate the control plane is up and running, the first Pod fails with the following:
```console
I0108 16:09:00.538760    1720 kubernetes_version.go:94] HTTP ERROR: Get https://kubernetes.default.svc/version: dial tcp: lookup kubernetes.default.svc on 10.96.0.10:53: no such host
W0108 16:09:00.546633    1720 util.go:315] 
Unable to find the programs kubectl or kubelet in the PATH.
These programs are used to determine which version of Kubernetes is running.
Make sure the /usr/bin directory is mapped to the container, 
either in the job.yaml file, or Docker command.

For job.yaml:
...
- name: usr-bin
  mountPath: /usr/bin
...

For docker command:
   docker -v $(which kubectl):/usr/bin/kubectl ....

Alternatively, you can specify the version with --version
   kube-bench --version <VERSION> ...

unable to determine benchmark version: Version check failed: unable to find the programs kubectl or kubelet in the PATH
Alternatively, you can specify the version with --version
```
Looks like `CoreDNS` is not ready or `DNS` completely setup for this Pod  by the time the code is attempting to access the Kubernetes API using `https://kubernetes.default.svc/version`

I have an idea on how to fix that, but I think it'll be better if I create another issue just for that so that we finish with this PR.

integration/integration.go
```diff
@@ -86,7 +86,7 @@ func getClientSet(configPath string) (*kubernetes.Clientset, error) {
 	return clientset, nil
 }
 
-func findPodForJob(clientset *kubernetes.Clientset, name string, tout, timer time.Duration) (*apiv1.Pod, error) {
+func findPodForJob(clientset *kubernetes.Clientset, name string, tout time.Duration) (*apiv1.Pod, error) {
```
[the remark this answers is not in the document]
robertojrojas: I renamed the parameter to `duration` , but we still need the `timeout` variable it outside the for loop so it doesn't reset on every iteration.

integration/integration.go
lizrice: You could do that under #551

integration/integration.go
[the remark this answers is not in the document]
lizrice: ah yes good point

integration/integration_test.go
```diff
@@ -12,59 +12,68 @@ import (
...
+		ctx.Delete()
+	}()
+
+	if err := loadImageFromDocker(*kubebenchImg, ctx); err != nil {
+		t.Fatalf("failed to load kube-bench image from Docker to KIND error: %v", err)
+	}
+
+	clientset, err := getClientSet(ctx.KubeConfigPath())
+	if err != nil {
+		t.Fatalf("failed to connect to Kuberntes cluster error: %v", err)
```
lizrice:
```suggestion
		t.Fatalf("failed to connect to Kubernetes cluster error: %v", err)
```

PR #556, opened 2020-01-07
Add config paths for microk8s

mkretz:
For running my own experiments I have added config paths for [Canonical's MicroK8s](https://github.com/ubuntu/microk8s). What do you think? Would it make sense to merge these to upstream?

### DIFF
--- a/cfg/config.yaml
+++ b/cfg/config.yaml
@@ -29,6 +29,7 @@ master:
       - /etc/kubernetes/manifests/kube-apiserver.yaml
       - /etc/kubernetes/manifests/kube-apiserver.manifest
       - /var/snap/kube-apiserver/current/args
+      - /var/snap/microk8s/current/args/kube-apiserver
     defaultconf: /etc/kubernetes/manifests/kube-apiserver.yaml
 
   scheduler:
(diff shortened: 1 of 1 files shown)


cfg/config.yaml
```diff
@@ -131,14 +140,17 @@ node:
       - "hyperkube kube-proxy"
       - "proxy"
     confs:
+      - /var/snap/microk8s/current/args/kube-proxy
```
lizrice: Please let's put this one as the last option in the list (they get checked in order)

lizrice:
hi @mkretz are you still interested in finishing this one off?

mkretz:
Hi @lizrice, sorry for the late reply. 

Absolutely! However, I think we should have someone from the [microk8s](https://github.com/ubuntu/microk8s) team review these paths to check whether any of them are likely to change in future releases. I have [opened an issue in that project](https://github.com/ubuntu/microk8s/issues/1001), asking whether one of the maintainers would have time for a short review. Ok with you?

ktsakalozos:
Thank you for doing this @mkretz. It gives us the motivation to harden MicroK8s. 

The paths are correct and they are not going to change in the near future.

lizrice:
Great so @mkretz if you wouldn't mind just moving the reference to `/var/snap/microk8s/current/args/kube-proxy` to be the last (rather than the first) in its list, we can merge this (per my review comment). Thanks!

mkretz:
Done, @lizrice ! I have also fixed the checks (yaml linting issue). Let me know if anything else needs to be done. Thanks for adopting these!

PR #561, opened 2020-01-09
Integration Tests: Adds generateDiff Function

### LINKED ISSUE #559: Integration tests: make it easier to see what has failed
In the event of failure, the output of `TestRunWithKind()` is a giant string of the expected and actual output, both of which are hundreds of lines long. To figure out the difference, I have ended up copying these strings into files, replacing the `\n` with newlines, and doing a file diff. Would be nice if the test output showed this diff automatically.

robertojrojas:
Integration tests:  Fixes #559
Adds `generateDiff` function to make it easier to see the difference between expected output and result.

Here is a sample output:
```console
--- FAIL: TestRunWithKind (160.86s)
    --- FAIL: TestRunWithKind/kube-bench (9.12s)
        integration_test.go:77: expected results
            
            Expected    (<)
            Result      (>)
            
            line: 25
            < [PASSa] 1.1.23 Ensure that the --service-account-lookup argument is set to true (Scored)
            > [PASS] 1.1.23 Ensure that the --service-account-lookup argument is set to true (Scored)
            line: 31
            < [bFAIL] 1.1.29 Ensure that the --client-ca-file argument is set as appropriate (Scored)
            > [FAIL] 1.1.29 Ensure that the --client-ca-file argument is set as appropriate (Scored)
            line: 516
            < 1 checks INFO2
            < 1 checks INFO3
            > [[NO MORE DATA]]
            
    --- PASS: TestRunWithKind/kube-bench-node (3.03s)
    --- PASS: TestRunWithKind/kube-bench-master (3.04s)
FAIL
FAIL    github.com/aquasecurity/kube-bench/integration  160.896s
FAIL
make: *** [integration-tests] Error 1
```

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -3,6 +3,8 @@
 package integration
 
 import (
+	"bufio"
+	"bytes"
 	"flag"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


integration/integration_test.go
```diff
@@ -72,8 +74,69 @@ func TestRunWithKind(t *testing.T) {
...
 			resultData = strings.TrimSpace(resultData)
 			if expectedData != resultData {
-				t.Errorf("expected: %q\n\n Got %q\n\n", expectedData, resultData)
+				t.Errorf("expected results\n\nExpected\t(<)\nResult\t(>)\n\n%s\n\n", generateDiff(expectedData, resultData))
 			}
 		})
 	}
 }
+
+func generateDiff(source, target string) string {
```
lizrice: I would add a comment to the effect that this is a simplistic diff that will be thrown off by the addition or removal of lines. (Full diff is pretty hard! I'd have used a library or shelled out to `diff` to get this done quickly. I'm OK with this approach though as it will help a lot in the normal scenario where an individual test gives us the wrong output.)

PR #562, opened 2020-01-10
Adds Retry Logic Until Timeout

### LINKED ISSUE #551: Integration tests: why do pods fail sometimes?

robertojrojas:
Adds Retry Logic Until Timeout.
Fixes #551

### DIFF
--- a/cmd/kubernetes_version.go
+++ b/cmd/kubernetes_version.go
@@ -4,17 +4,18 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/golang/glog"
 )
 
 func getKubeVersionFromRESTAPI() (string, error) {
-	k8sVersionURL := getKubernetesURL()
 	serviceaccount := "/var/run/secrets/kubernetes.io/serviceaccount"
 	cacertfile := fmt.Sprintf("%s/ca.crt", serviceaccount)
 	tokenfile := fmt.Sprintf("%s/token", serviceaccount)
@@ -30,7 +31,9 @@ func getKubeVersionFromRESTAPI() (string, error) {
 	}
 	token := strings.TrimSpace(string(tb))
 
-	data, err := getWebData(k8sVersionURL, token, tlsCert)
+	timeDuration := 30 * time.Second
+	waitDuration := 1 * time.Second
+	data, err := getWebDataRetry(getKubernetesURLs(), token, tlsCert, timeDuration, waitDuration)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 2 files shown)


cmd/kubernetes_version.go
```diff
@@ -122,21 +148,9 @@ func loadCertficate(certFile string) (*tls.Certificate, error) {
...
-		k8sPort := os.Getenv("KUBERNETES_SERVICE_PORT_HTTPS")
-		if !isEmpty(k8sHost) && !isEmpty(k8sPort) {
-			return fmt.Sprintf("https://%s:%s/version", k8sHost, k8sPort)
-		}
-
-		glog.V(2).Info(fmt.Sprintf("KUBE_BENCH_K8S_ENV is set, but environment variables KUBERNETES_SERVICE_HOST or KUBERNETES_SERVICE_PORT_HTTPS are not set"))
+func getKubernetesURLs() []string {
+	return []string{
+		"https://kubernetes.default.svc/version",
+		fmt.Sprintf("https://%s:%s/version", os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT_HTTPS")),
```
lizrice: This doesn't use the environment variables if they were set, unless `kubernetes.default.svc` fails. If the env vars are set, that's what kube-bench should use, so I don't think there's any reason to change this function or to have it return a list. 

Also this is no longer checking `KUBE_BENCH_K8S_ENV` and the tests below seem to remove it. If you are retiring the use of some env vars I think you should document that in the comments about the PR so that whoever is reviewing it understands why you're doing it.

cmd/kubernetes_version.go
```diff
@@ -69,6 +72,30 @@ func extractVersion(data []byte) (string, error) {
...
+func getWebDataRetry(srvURLs []string, token string, cacert *tls.Certificate, timeoutDuration, waitDuration time.Duration) (data []byte, err error) {
+	timeout := time.After(timeoutDuration)
+	wait := time.Duration(waitDuration)
+	for {
+		select {
+		case <-timeout:
+			err = errors.New("timeout - getting Kubernetes version using REST API")
+			return
+		default:
+			for _, srvURL := range srvURLs {
```
lizrice: Per comment below, I don't think this should be a list. It should either be set by env vars, or it should be the default value.

cmd/kubernetes_version.go
```diff
@@ -69,6 +72,30 @@ func extractVersion(data []byte) (string, error) {
...
+				data, err = getWebData(srvURL, token, cacert)
+				if err == nil {
+					return
+				}
+				glog.V(2).Info(fmt.Sprintf("HTTP ERROR: %v\n", err))
+				continue
```
lizrice: given that you're at the end of the loop, what is `continue` intended to do here?

cmd/kubernetes_version.go
```diff
@@ -69,6 +72,30 @@ func extractVersion(data []byte) (string, error) {
...
+			}
+			glog.V(2).Info(fmt.Sprintf("Waiting (%s) to retry getting Kubernetes version using REST API", wait))
+			time.Sleep(wait)
+			continue
```
lizrice: same question about this `continue`

robertojrojas:
Closing in favor of a better approach.

Issue #541, opened 2019-12-12
CIS 1.5 file permissions are set to 644 or more restrictive


IceBear2k:
In CIS 1.5 1.1.7 (for example) states: "Ensure that the etcd pod specification file permissions are set to 644 or more restrictive (Scored)"

A lot of checks (1.1.7 included) only check for very specific permissions:
https://github.com/aquasecurity/kube-bench/blob/master/cfg/cis-1.5/master.yaml#L140

Here, only '644', '640' and '600' are specifically checked. This results in file permissions '440' (for example) failing, even though they match the specification "set to 644 or more restrictive".

Expected behavior:
444, 440, 400, 000 should also pass these checks.

PR #565, opened 2020-01-16
Add option to do bitmask 

yoavrotems:
Bitwise AND between two value in order to compare file permissions
solved issue https://github.com/aquasecurity/kube-bench/issues/541.
According to the new format, changed all the yaml configuration files to comply


check/test_test.go
```diff
@@ -690,4 +703,14 @@ func TestToNumeric(t *testing.T) {
 			t.Errorf("TestToNumeric - Expected to return %d,%d , but instead got %d,%d", 5, 6, f, s)
 		}
 	}
+	for _, c := range cases {
+		f, s, err := toNumericOctal(c.firstValue, c.secondValue)
+		if c.expectedToFail && err == nil {
```
robertojrojas: I think it's a best practice for **Unit tests** to test only one function at a time, so the test for `toNumericOctal` belongs on its own Test function **`TestToNumericOctal`**.

lizrice: Or even better, put the bitmask in octal representation (Yehuda’s suggestion in our tram meeting) and then you wouldn’t need a conversion function

lizrice:
Oh, wait - I think you can remove the `toNumericOctal()` function and its associated tests altogether now, right?

PR #567, opened 2020-01-21
FIX : Resolved bug in related to apiserverbin variable name in check 1.2.21

mustafa-rean:
Resolved bug in **master.yml** for **cis-1.5** for the **$apiserverbin** variable name in check **1.2.21**
For check **1.2.21**, a wrong variable name is referenced **$apiserver** which actually should be **$apiserverbin**
This was making check to fail even if the **profiling** argument is set to **false** for **apiserver**
Also resolved remediation description which din't list all the required **tls-cipher-suites** which are actually checked in the check **1.2.35**

### DIFF
--- a/cfg/cis-1.5/master.yaml
+++ b/cfg/cis-1.5/master.yaml
@@ -734,7 +734,7 @@ groups:
 
       - id: 1.2.21
         text: "Ensure that the --profiling argument is set to false (Scored)"
-        audit: "/bin/ps -ef | grep $apiserver | grep -v grep"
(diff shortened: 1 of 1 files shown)


mustafa-rean:
Welcome @lizrice. 
Please let me know what will be expected time for this PR to be merged and released as we need this fix in our workflow sooner :)

lizrice:
We have just cut a release now @mustafa-rean: https://github.com/aquasecurity/kube-bench/releases/tag/v0.2.3

mustafa-rean:
Thanks for the updates @lizrice

PR #569, opened 2020-01-27
Set all host-mounted volumes to be read-only.

nickrmc83:
By setting all host-mounted volumes to be read-only we reduce the likelihood
any host filesystem is modified by running kube-bench. This is a defense in depth measure that improves confidence in running kube-bench in production environments as well as allowing kube-bench to run where restrictive pod security policies are enforced.


robertojrojas:
@nickrmc83 thanks for your contribution!

robertojrojas:
LGTM

PR #570, opened 2020-01-31
CIS GKE 1.0.0 benchmark

ttousai:


ttousai:
@lizrice @robertojrojas in section 6 (managed services) we can only audit the recommendations if the gcloud command line  tool is installed. I am not sure this is the case. The implication is that section 6 controls will have to be set to `type: manual` since it is mostly likely going to be run on the k8s operators workstation. 

WDYT?

robertojrojas:
> @lizrice @robertojrojas in section 6 (managed services) we can only audit the recommendations if the gcloud command line tool is installed. I am not sure this is the case. The implication is that section 6 controls will have to be set to `type: manual` since it is mostly likely going to be run on the k8s operators workstation.
> 
> WDYT?

@ttousai  Setting the type as `manual` for this seems fine to me.  IMHO, `kube-bench` should stay as tool agnostic as possible.

lizrice:
> > @lizrice @robertojrojas in section 6 (managed services) we can only audit the recommendations if the gcloud command line tool is installed. I am not sure this is the case. The implication is that section 6 controls will have to be set to `type: manual` since it is mostly likely going to be run on the k8s operators workstation.
> > WDYT?
> 
> @ttousai Setting the type as `manual` for this seems fine to me. IMHO, `kube-bench` should stay as tool agnostic as possible.

I am fine with this at least for an initial release. It may turn out that anyone who wants to run the GKE benchmarks has the tool installed, but we can come back to it based on user feedback.

lizrice:
@ttousai just a ping to see if you could do the README update please?

ttousai:
@lizrice @robertojrojas I have pushed changes including the README. Please take a look.

Issue #572, opened 2020-02-06
target validation has hardcoded benchmark versions


leodotcloud:
```
kube-bench --benchmark rke-cis-1.5 run --targets master --scored --nosummary --noremediations --v=5  --json

The specified --targets "master" does not apply to the CIS Benchmark rke-cis-1.5 \n Valid targets []
```

In the past, we copied cis-1.4 to rke-cis-1.4, modify the tests and run specifying the version as `rke-cis-1.4`, but this behavior is broken in the latest version of kube-bench due to the validation check using the hardcoded versions. https://github.com/aquasecurity/kube-bench/blob/v0.2.3/cmd/common.go#L389

Can we have a CLI toggle to disable this validation check? If this is acceptable, I can make a quick PR.

leodotcloud:
I have figured a better/cleaner way, specify these targets in config.yaml following the pattern of the rest of the config. Will make a PR in few minutes.

PR #573, opened 2020-02-06
move target mapping to config.yaml

leodotcloud:
Addresses: https://github.com/aquasecurity/kube-bench/issues/572

Instead of having the mapping hard coded in code, host it in the config.yaml so that others can manage it without having to rebuild the code.


cmd/common.go
```diff
@@ -287,6 +287,15 @@ func loadVersionMapping(v *viper.Viper) (map[string]string, error) {
 	return kubeToBenchmarkMap, nil
 }
 
+func loadTargetMapping(v *viper.Viper) (map[string][]string, error) {
+	benchmarkVersionToTargetsMap := v.GetStringMapStringSlice("target_mapping")
+	if benchmarkVersionToTargetsMap == nil || (len(benchmarkVersionToTargetsMap) == 0) {
```
robertojrojas: No need to check for **`nil`** on maps. The **`len`** check should suffice.

cmd/common.go
```diff
@@ -378,18 +387,16 @@ func PrintOutput(output string, outputFile string) {
...
-	"cis-1.3": []string{string(check.MASTER), string(check.NODE)},
-	"cis-1.4": []string{string(check.MASTER), string(check.NODE)},
-	"cis-1.5": []string{string(check.MASTER), string(check.NODE), string(check.CONTROLPLANE), string(check.ETCD), string(check.POLICIES)},
-}
-
 // validTargets helps determine if the targets
 // are legitimate for the benchmarkVersion.
-func validTargets(benchmarkVersion string, targets []string) bool {
+func validTargets(benchmarkVersion string, targets []string, v *viper.Viper) (bool, error) {
+	benchmarkVersionToTargetsMap, err := loadTargetMapping(v)
```
robertojrojas: I think it would be better if we load the `benchmarkVersionToTargetsMap` from the config file just once, keep it in a package level variable and then just reuse it. 
This could also help on those places calling `validTargets` so they don't have to check for the same error over and over again.

cfg/config.yaml
```diff
@@ -173,3 +173,17 @@ version_mapping:
...
+    - "node"
+  "cis-1.4":
+    - "master"
+    - "node"
+  "cis-1.5":
+    - "master"
+    - "node"
+    - "controlplane"
+    - "etcd"
+    - "policies"
```
lizrice: I think there also needs to be an entry for "rh-0.7" which lists "master" and "node"

lizrice: Hi @leodotcloud, if I'm not mistaken it's just this missing `rh-0.7` entry that's standing between us merging this PR, would you be OK to add that?

leodotcloud: Sorry, have been busy. Will try to rework on this PR early next week.

cmd/common.go
```diff
@@ -426,6 +426,13 @@ func PrintOutput(output string, outputFile string) {
 	}
 }
 
+var benchmarkVersionToTargetsMap = map[string][]string{
+	"cis-1.3": []string{string(check.MASTER), string(check.NODE)},
+	"cis-1.4": []string{string(check.MASTER), string(check.NODE)},
+	"cis-1.5": []string{string(check.MASTER), string(check.NODE), string(check.CONTROLPLANE), string(check.ETCD), string(check.POLICIES)},
+	"gke-1.0": []string{string(check.MASTER), string(check.NODE), string(check.CONTROLPLANE), string(check.ETCD), string(check.POLICIES), string(check.MANAGEDSERVICES)},
+}
```
lizrice, line 434: We also need targets for eks-1.0 now

yoavrotems:
Close this due to addressing it in #682

PR #576, opened 2020-02-11
Fixes issue #574: specifying absolute path for grep and sed

### LINKED ISSUE #574: Running in a kubernetes cluster don't work as expected when check cis-1.4-1.1.8
**k8s version**: 1.14.8
**kube-bench version**: latest (0.2.3)
**run command**: wget https://github.com/aquasecurity/kube-bench/blob/master/job-master.yaml && kubectl apply -f job-master.yaml
**problem**:  `--profiling=false` is exists but this check is not passed

https://github.com/aquasecurity/kube-bench/blob/77f66511e70c4b2c1f83988786e3de04ec7349db/cfg/cis-1.4/master.yaml#L120-L134

debug in k8s container:

图片

```
/opt/kube-bench # kube-bench version
v0.2.3
/opt/kube-bench # kube-bench |grep 1.1.8
[FAIL] 1.1.8 Ensure that the --profiling argument is set to false (Scored)
1.1.8 Edit the API server pod specification file /etc/kubernetes/manifests/kube-apiserver.yaml
/opt/kube-bench # /bin/ps -ef | grep apiserver | grep -v grep
root      3334  3305  2 Feb05 ?        01:10:07 kube-apiserver --audit-log-maxbackup=10 --audit-log-maxsize=100 --audit-log-path=/var/log/kubernetes/kubernetes.audit --audit-log-maxage=7 --audit-policy-file=/etc/kubernetes/audit-policy.yml --apiserver-count=500 --endpoint-reconciler-type=lease --enable-aggregator-routing=true --runtime-config=admissionregistration.k8s.io/v1beta1 --profiling=false --advertise-address=...
```

When run kube-bench bin file in master node it check passed, but run in a k8s container it check failed

BTW, other check itemes which check it's flag are failed too(there are needed flag value but check failed like 1.1.8)

mozillazg:
Fixes #574

### DIFF
--- a/cfg/cis-1.5/master.yaml
+++ b/cfg/cis-1.5/master.yaml
@@ -282,7 +282,7 @@ groups:
 
       - id: 1.1.11
         text: "Ensure that the etcd data directory permissions are set to 700 or more restrictive (Scored)"
-        audit: /bin/ps -ef | /bin/grep $etcdbin | /bin/grep -- --data-dir | sed 's%.*data-dir[= ]\([^ ]*\).*%\1%' | xargs stat -c %a
+        audit: /bin/ps -ef | /bin/grep $etcdbin | /bin/grep -- --data-dir | /bin/sed 's%.*data-dir[= ]\([^ ]*\).*%\1%' | xargs stat -c %a
         tests:
           test_items:
             - flag: "700"
@@ -299,7 +299,7 @@ groups:
 
       - id: 1.1.12
         text: "Ensure that the etcd data directory ownership is set to etcd:etcd (Scored)"
-        audit: /bin/ps -ef | /bin/grep $etcdbin | /bin/grep -- --data-dir | sed 's%.*data-dir[= ]\([^ ]*\).*%\1%' | xargs stat -c %U:%G
+        audit: /bin/ps -ef | /bin/grep $etcdbin | /bin/grep -- --data-dir | /bin/sed 's%.*data-dir[= ]\([^ ]*\).*%\1%' | xargs stat -c %U:%G
         tests:
           test_items:
             - flag: "etcd:etcd"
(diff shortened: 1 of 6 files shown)


cfg/cis-1.5/master.yaml
```diff
@@ -282,7 +282,7 @@ groups:
 
       - id: 1.1.11
         text: "Ensure that the etcd data directory permissions are set to 700 or more restrictive (Scored)"
-        audit: ps -ef | grep $etcdbin | grep -- --data-dir | sed 's%.*data-dir[= ]\([^ ]*\).*%\1%' | xargs stat -c %a
+        audit: /bin/ps -ef | /bin/grep $etcdbin | /bin/grep -- --data-dir | /bin/sed 's%.*data-dir[= ]\([^ ]*\).*%\1%' | xargs stat -c %a
```
mozillazg: `xargs` has same problem but there isn't `/bin/xargs` in alpine(it's in `/usr/bin` but overwrite by mount from host). Maybe we should change the mount dir like still mount `/usr/bin` from host but use `/usr/local/sbin` inside container, use this way we can both use `kubectl` command from host and use all builtin commands from alpine.

cfg/rh-0.7/master.yaml
```diff
@@ -621,7 +621,7 @@ groups:
 
       - id: 1.34
         text: "Set the encryption provider to aescbc for etcd data at rest"
-        audit: "grep -A1 experimental-encryption-provider-config /etc/origin/master/master-config.yaml | sed -n '2p' | awk '{ print $2 }' | xargs grep -A1 providers"
+        audit: "/bin/grep -A1 experimental-encryption-provider-config /etc/origin/master/master-config.yaml | /bin/sed -n '2p' | awk '{ print $2 }' | xargs /bin/grep -A1 providers"
```
mozillazg: `awk` has same problem too.

mozillazg:
Use absolute path can't fix same problem of `xargs` and `awk`.

mozillazg:
Use #577 instead.

PR #579, opened 2020-02-12
Support parse boolean flag with no value

### LINKED ISSUE #578: Check boolean flag which has no value didn't work as expected
The boolean flag can be set with no value, in this case it's value should be parsed as `true`, but in current it's parsed as other value.

For example:

The boolean flag `--peer-client-cert-auth` of etcd can be set with `--peer-client-cert-auth` or `--peer-client-cert-auth=true`, but the parsed result is different:

* `--peer-client-cert-auth` -> `--peer-client-cert-auth` (it should be parsed as `true`).
* `--peer-client-cert-auth=true` -> `true`

mozillazg:
Fixes #578

### DIFF
--- a/check/data
+++ b/check/data
@@ -308,6 +308,16 @@ groups:
               value: '^1\.12.*$'
             set: true
 
+    - id: 27
+      text: "check boolean flag with no value"
+      tests:
+        test_items:
+          - flag: "--peer-client-cert-auth"
+            compare:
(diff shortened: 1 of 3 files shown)


mozillazg:
@robertojrojas When the users provides `--peer-client-cert-auth=false`, it's value will be parsed as `false`. I have added a test for this case.

PR #581, opened 2020-02-12
Improve the check of `--audit-log-maxage` to accept appropriate number

### LINKED ISSUE #580: The check of cis-1.5-1.2.23 and cis-1.4-1.1.16 should accept any number
https://github.com/aquasecurity/kube-bench/blob/77f66511e70c4b2c1f83988786e3de04ec7349db/cfg/cis-1.4/master.yaml#L257-L266
the current check limit the value of `--audit-log-maxage` must >= 30, but in CIS it accept any number > 0:

> set the --audit-log-maxage parameter to 30 **or as an appropriate number of days**

图片

mozillazg:
Fixes #580

### DIFF
--- a/cfg/cis-1.3/master.yaml
+++ b/cfg/cis-1.3/master.yaml
@@ -260,8 +260,8 @@ groups:
           test_items:
             - flag: "--audit-log-maxage"
               compare:
-                op: gte
-                value: 30
+                op: gt
+                value: 0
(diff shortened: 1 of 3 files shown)


lizrice:
Per my comment in #580 I don't think this is really in the spirit of the benchmark, so I don't think we should accept this change unfortunately.

PR #593, opened 2020-03-03
Get Kubernetes Version: Adds Retry Logic 

### LINKED ISSUE #551: Integration tests: why do pods fail sometimes?

robertojrojas:
When getting the Kubernetes version this code uses a retry logic in case errors like DNS not ready happen.
 
Closes #551

### DIFF
--- a/cmd/kubernetes_version.go
+++ b/cmd/kubernetes_version.go
@@ -43,8 +43,12 @@ func getKubeVersionFromRESTAPI() (string, error) {
 	return k8sVersion, nil
 }
 
+// The idea of this function is so if Kubernetes DNS is not completely seetup and the
+// Container where kube-bench is running needs time for DNS configure.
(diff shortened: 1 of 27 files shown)


cmd/kubernetes_version.go
```diff
@@ -42,6 +43,20 @@ func getKubeVersionFromRESTAPI() (string, error) {
 	return k8sVersion, nil
 }
 
+func getWebDataWithRetry(k8sVersionURL, token string, cacert *tls.Certificate) (data []byte, err error) {
+	tries := 0
+	for tries < 10 {
```
lizrice:
```suggestion
      // We retry a few times in case the DNS service has not had time to come up
	for tries < 10 {
```

Suggest adding a comment so that we know why it's an arbitrary number of retries, and in case we ever need to increase that number

PR #601, opened 2020-03-26
fix the default make to use BUILD_OS instead of static TARGET_OS

yashbhutwala:
- [X] on platforms other than linux, default make would not work


lizrice:
Thanks for getting involved but I don't think this fix is right. In general we need to build kube-bench to run on Linux (that's the TARGET_OS). It doesn't really make sense to run kube-bench on other OSs, because all the file locations and so-on are Linux-based. 

We can run kube-bench within, say, kind. kind runs in a Linux container, so kube-bench still needs to be built for Linux regardless of the host OS. 

The BUILD_OS is used to download the right version of stern, which runs on the host OS.

PR #609, opened 2020-04-18
Add integration tests for cis 1.3 and cis 1.5

### LINKED ISSUE #606: Add integration tests for cis 1.3 and cis 1.5
It's useful when we want to do refactoring for some codes.

mozillazg:
Closes #606


integration/testdata/cis-1.3/add-tls-kind-k8s112.yaml
```diff
@@ -0,0 +1,19 @@
...
+  apiServerAddress: "0.0.0.0"
+
+kubeadmConfigPatchesJson6902:
+  - group: kubelet.config.k8s.io
+    version: v1beta1
+    kind: KubeletConfiguration
+    patch: |
+      - op: add
+        path: /tlsCipherSuites
+        value: ["TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256","TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"]
```
lizrice, line 14: I don't know why we have to do this patch, but since we have it in the current test I am OK with keeping it in the new ones

integration/integration_test.go
```diff
@@ -16,7 +16,7 @@ import (
 var kubebenchImg = flag.String("kubebenchImg", "aquasec/kube-bench:latest", "kube-bench image used as part of this test")
 var timeout = flag.Duration("timeout", 10*time.Minute, "Test Timeout")
 
-func TestRunWithKind(t *testing.T) {
+func TestCheckCIS13WithKind(t *testing.T) {
```
lizrice, line 19: I see the point of having separate test functions for the different clusters, so they can easily be run independently. But it seems like a lot of repeated code. The part that actually sets up the cluster and runs the test could be pulled into a separate function that each of the three test function calls

mozillazg: Fixed.

PR #610, opened 2020-04-19
Run audit as shell script instead of as single line command

### LINKED ISSUE #604: Support run tests which have complex shell commands
There are some scored item have been marked `manual` because of it need complex shell commands to run test but kube-bench don't support it. Change to support run complex shell commands can  avoid this issue and fix issues like https://github.com/aquasecurity/kube-bench/issues/598 too.

kube-bench run tests via parse test command as a simple single line command then run it, this way don't support complex shell commands.
https://github.com/aquasecurity/kube-bench/blob/7cd6b32ebb5f6c9503daf6069fe8480e41020505/check/check.go#L206

What about change to run the test command just as run a shell script, this way can support complex commands.

mozillazg:
Fixes #604


lizrice:
Sorry for being slow on this. From a very quick glance it looks similar to what @yoavrotems did with https://github.com/aquasecurity/bench-common/pull/87

mozillazg:
@lizrice Should i close this PR or use  similar codes from aquasecurity/bench-common#87 ?

mozillazg:
@yoavrotems ping

PR #615, opened 2020-04-29
Correct typo

gliptak:


gliptak:
@lizrice @aquasecurity please review

lizrice:
Duplicate of #616, I think

gliptak:
@lizrice not a duplicate of #616 while filename is same, path is different

PR #630, opened 2020-06-24
Refactor: remove ContinueWithError

yoavrotems:
Remove Continue with error function

### DIFF
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -70,8 +70,7 @@ func ps(proc string) string {
 	cmd := exec.Command("/bin/ps", "-C", proc, "-o", "cmd", "--no-headers")
 	out, err := cmd.Output()
 	if err != nil {
-		glog.V(2).Info(err)
-		fmt.Fprintf(os.Stderr, "%s: %s", cmd.Args, err)
+		glog.V(2).Info(fmt.Errorf("%s: %s", cmd.Args, err))
 	}
 
 	glog.V(2).Info(fmt.Sprintf("ps - returning: %q", string(out)))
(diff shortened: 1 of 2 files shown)


yoavrotems:
Tests passed 
https://travis-ci.org/github/aquasecurity/kube-bench/jobs/701672953
https://travis-ci.org/github/aquasecurity/kube-bench/builds/701672987
But for some reason status here is stuck on Pending

cmd/util.go
```diff
@@ -82,7 +70,8 @@ func ps(proc string) string {
 	cmd := exec.Command("/bin/ps", "-C", proc, "-o", "cmd", "--no-headers")
 	out, err := cmd.Output()
 	if err != nil {
-		continueWithError(fmt.Errorf("%s: %s", cmd.Args, err), "")
+		glog.V(2).Info(err)
+		glog.V(2).Info(fmt.Errorf("%s: %s", cmd.Args, err))
```
lizrice, line 74: These logs will only show up if error logging is level 2 or higher; also logs might be going somewhere else like to a file rather than stdout or stderr. So I think we should stick with the original behaviour of printing the message to stderr.

```suggestion
		glog.V(2).Info(err)
		fmt.Fprintf(os.Stderr, "%s: %s", cmd.Args, err))
```

lizrice:
@yoavrotems Please can you revisit the name of this PR? Something like "Refactor: remove ContinueWithError" would be more meaningful to someone else looking at the PR list

yoavrotems:
@lizrice I changed but apparently didn't apply my change...

Issue #635, opened 2020-07-05
Take a time to checking if there is any scored check which hasn't an audit command when possible


mozillazg:
There are some scored checks not have an audit command when it is possible been check via command. For example:

https://github.com/aquasecurity/kube-bench/blob/1b5b6c2afe5fb1e0060ae96346f77484842cef75/cfg/cis-1.5/node.yaml#L100-L106

And some scored checks have been marked as `manual` even if it has an audit command. For example:

https://github.com/aquasecurity/kube-bench/blob/1b5b6c2afe5fb1e0060ae96346f77484842cef75/cfg/cis-1.5/master.yaml#L288-L296

lizrice:
Fixed with #640 and #641, but please open a new issue if you find any more tests incorrectly marked as manual. Thanks!

PR #641, opened 2020-07-11
Add tests for 1.1.19、1.1.20 and 1.1.21 of cis-1.5

mozillazg:
#635


cfg/cis-1.5/master.yaml
```diff
@@ -297,8 +303,22 @@ groups:
 
       - id: 1.1.20
         text: "Ensure that the Kubernetes PKI certificate file permissions are set to 644 or more restrictive (Scored)"
-        audit: "stat -c %n\ %a /etc/kubernetes/pki/*.crt"
-        type: "manual"
+        audit: |
```
yoavrotems, line 306: I think It will better to use bitmask flag here and  multiple values.
It will be more accurate

cfg/cis-1.5/master.yaml
```diff
@@ -307,8 +327,22 @@ groups:
 
       - id: 1.1.21
         text: "Ensure that the Kubernetes PKI key file permissions are set to 600 (Scored)"
-        audit: "stat -c %n\ %a /etc/kubernetes/pki/*.key"
-        type: "manual"
+        audit: |
```
yoavrotems, line 330: ditto

mozillazg:
@yoavrotems Thanks for your review! I have fixed it, please review again. Thanks!

yoavrotems:
Sorry it seems I mislead you,
The problem with summing all the permission up is that some permission could cover for other, for example if you the cert that has   220, and one with 777, it will still pass the test.
What I thought is that multiple values is something we have in kube-bench, but it in bench-common https://github.com/aquasecurity/bench-common/blob/master/check/check.go#L105 .
This functionality could take multiple line result and check that every line is passing our test, then using the bitmask on each file will work.
You are more then welcome to open a PR about it, currently I opened an issue about it https://github.com/aquasecurity/kube-bench/issues/647.
Before this will be made I can't approve your changes.
And once again thanks for contributing

yoavrotems:
Good job with the https://github.com/aquasecurity/kube-bench/pull/652! now this PR need to use the multiple values and that it

mozillazg:
@yoavrotems ready for review

cfg/cis-1.5/master.yaml
```diff
@@ -297,8 +304,15 @@ groups:
 
       - id: 1.1.20
         text: "Ensure that the Kubernetes PKI certificate file permissions are set to 644 or more restrictive (Scored)"
-        audit: "stat -c %n\ %a /etc/kubernetes/pki/*.crt"
-        type: "manual"
+        audit: "stat -c permissions=%a /etc/kubernetes/pki/*.crt"
```
yoavrotems, line 307: Hey, I tested it and notice it will not check sub directories, better use find, like the first test you changed 1.1.19 
It also the audit commend in the new CIS version 
image

cfg/cis-1.5/master.yaml
```diff
@@ -307,8 +321,15 @@ groups:
 
       - id: 1.1.21
         text: "Ensure that the Kubernetes PKI key file permissions are set to 600 (Scored)"
-        audit: "stat -c %n\ %a /etc/kubernetes/pki/*.key"
-        type: "manual"
+        audit: "stat -c permissions=%a /etc/kubernetes/pki/*.key"
```
yoavrotems, line 324: ditto, better use find, beside the audit command the use of multiple values and bitmask is great :)

cfg/cis-1.5/master.yaml
mozillazg: Looks like the new CIS version which you mean is cis 1.5.1, but this PR is talk about cis 1.5, maybe we should follow the cis 1.5 instead of cis 1.5.1?

PR #648, opened 2020-07-22
Fix supported targets for EKS benchmark

thehunt33r:
Hello!

We've been using kube-bench for quite a while in our CI pipeline for our EKS clusters. Up until the release of the eks benchmark, we had been using the vanilla 1.17 benchmark.

Following AWS's announcement of the new EKS-specific benchmark, I set out to try it out and switch to it. However I encountered errors while doing that.

The supported targets for the EKS benchmark weren't matching the documentation, or the config files for the benchmark.

```
kube-bench --benchmark eks-1.0 run --targets managedservice
The specified --targets "managedservice" does not apply to the CIS Benchmark eks-1.0 \n Valid targets [
```

This PR fixes the supported targets for EKS benchmark:

* Before that, `eks-1.0` was supposed to be supporting node, policies and managedservices. However using policies or managedservices as a target would result in an error.
  * The `benchmarkVersionToTargetsMap` was updated so that the `eks-1.0` benchmark would reflect the documentation
* The tests for the `validTargets` method weren't testing the `eks-1.0` benchmark. This has been added.
* In the config files for the benchmarks, there are also checks for `controlplane` for `eks-1.0`. This has therefore been added as a valid target, and documented

Please let me know if anything's missing.

Cheers,

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,6 @@ Tests are configured with YAML files, making this tool easy to update as test sp
 Table of Contents
 =================
 
-- [Table of Contents](#table-of-contents)
   - [CIS Kubernetes Benchmark support](#cis-kubernetes-benchmark-support)
   - [Installation](#installation)
   - [Running kube-bench](#running-kube-bench)
--- a/README.md
+++ b/README.md
@@ -259,7 +259,7 @@ go build -o kube-bench .
 ./kube-bench
 ```
 
-## Running on OpenShift
+### Running on OpenShift
 
 | OpenShift Hardening Guide | kube-bench config |
 |---|---|
--- a/README.md
+++ b/README.md
@@ -8,13 +8,13 @@
 [cov]: https://codecov.io/github/aquasecurity/kube-bench
 <img src="images/kube-bench.png" width="200" alt="kube-bench logo">
 
-kube-bench is a Go application that checks whether Kubernetes is deployed securely by running the checks documented in the [CIS Kubernetes Benchmark](https://www.cisecurity.org/benchmark/kubernetes/). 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -26,32 +26,34 @@ Tests are configured with YAML files, making this tool easy to update as test sp
...
-  * [Omitting checks](#omitting-checks)
-* [Roadmap](#roadmap)
-* [Testing locally with kind](#testing-locally-with-kind)
-* [Contributing](#contributing)
-  * [Bugs](#bugs)
-  * [Features](#features)
-  * [Pull Requests](#pull-requests)
-
-      
+- [Table of Contents](#table-of-contents)
```
lizrice, line 29: There's already a Table of contents heading above so I don't think this is necessary, and it introduces an extra level of indentation for all the bullets below
```suggestion
```

README.md
```diff
@@ -257,7 +259,7 @@ go build -o kube-bench .
 ./kube-bench
 ```
 
-## Running on OpenShift 
+## Running on OpenShift
```
lizrice, line 262: Hmm, this is somehow at the wrong heading level 
```suggestion
### Running on OpenShift
```

thehunt33r:
Wonderful thank you @lizrice! 

It appears the CI build is broken

```bash
$ gem install --no-ri --no-rdoc fpm
ERROR:  While executing gem ... (OptionParser::InvalidOption)
    invalid option: --no-ri
The command "gem install --no-ri --no-rdoc fpm" failed and exited with 1 during .
```

I've pushed a change to replace `gem install --no-ri --no-rdoc fpm` by `gem install --no-document fpm` as documented in the [release notes](https://www.ruby-lang.org/en/news/2018/12/25/ruby-2-6-0-released/) for Ruby 2.6.

Issue #655, opened 2020-08-03
Easily-missed bad version check


lizrice:
If kube-bench is using `kubectl` to auto-detect the running version of kubernetes, but the kubeconfig isn't available, the version check will fail and kube-bench falls back to a default of kubernetes 1.11. 

I found this easily by running kubernetes on a vagrant machine. The kubeconfig file was installed for the user called vagrant, but running `kube-bench` as root meant kubeconfig was not set up. 

In this situation `-v 3 --logtostderr`  includes these logs:

```
I0803 17:05:45.584195   12442 util.go:53] Client Version: v1.18.3
The connection to the server localhost:8080 was refused - did you specify the right host or port?
I0803 17:05:45.584503   12442 util.go:351] Unable to get Kubernetes version from kubectl, using default version: 1.11
```

We should update the default version to a current version (say, 1.18) rather than an ancient version of kubernetes that is no longer supported

We should emit a warning in this situation to suggest specifying the version or ensuring that kubeconfig is available.

yoavrotems:
PRs #656 and https://github.com/aquasecurity/kube-bench/pull/657 solves this issue

PR #656, opened 2020-08-03
Warn if kubectl can't autodetect the version 

lizrice:
See #655. This outputs a warning if kubectl doesn't have access to kubeconfig and can't autodetect the version. 

Modified code to only run `getBenchmarkVersion()` once per command, so that this new warning doesn't get emitted four times! (Note that the `loadConfig()` function can get called several times from the code in root.go) 

This PR doesn't update the default version yet so it's not a full fix for the issue.

### DIFF
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -276,7 +276,7 @@ func multiWordReplace(s string, subname string, sub string) string {
 const missingKubectlKubeletMessage = `
 Unable to find the programs kubectl or kubelet in the PATH.
 These programs are used to determine which version of Kubernetes is running.
-Make sure the /usr/local/mount-from-host/bin directory is mapped to the container, 
+Make sure the /usr/local/mount-from-host/bin directory is mapped to the container,
 either in the job.yaml file, or Docker command.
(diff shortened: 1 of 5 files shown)


cmd/util.go
```diff
@@ -346,6 +346,9 @@ func getVersionFromKubectlOutput(s string) string {
 	serverVersionRe := regexp.MustCompile(`Server Version: v(\d+.\d+)`)
 	subs := serverVersionRe.FindStringSubmatch(s)
 	if len(subs) < 2 {
+		if strings.Contains(s, "The connection to the server") {
+			continueWithError(nil, `Warning: Kubernetes version was not auto-detected because kubectl could not connect to the Kubernetes server. This may be because the kubeconfig information is missing or has credentials that do not match the server.`)
```
yoavrotems, line 350: I think that it will be better to just print it to stderr since we are refactoring continueWithError function https://github.com/aquasecurity/kube-bench/pull/630

lizrice: good point!

Issue #658, opened 2020-08-04
Make testitem-set to be default to true


yoavrotems:
In almost every test we have "set: true".
With making it default true the yaml will be neater

PR #668, opened 2020-08-10
Refactor testitem-set

yoavrotems:
Refactor testitem-set to be default true https://github.com/aquasecurity/kube-bench/issues/658


check/test.go
```diff
@@ -398,3 +398,16 @@ func toNumeric(a, b string) (c, d int, err error) {
 
 	return c, d, nil
 }
+
+func (t *testItem) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type buildTest testItem
+
+	// Make Set parameter to be treu by default.
```
lizrice, line 405:
```suggestion
	// Make Set parameter to be true by default.
```

PR #678, opened 2020-08-13
Add cis 1.6

yoavrotems:


cmd/common_test.go
```diff
@@ -211,9 +211,9 @@ func TestMapToCISVersion(t *testing.T) {
 		{kubeVersion: "1.13", succeed: true, exp: "cis-1.4"},
 		{kubeVersion: "1.14", succeed: true, exp: "cis-1.4"},
 		{kubeVersion: "1.15", succeed: true, exp: "cis-1.5"},
-		{kubeVersion: "1.16", succeed: true, exp: "cis-1.5"},
-		{kubeVersion: "1.17", succeed: true, exp: "cis-1.5"},
-		{kubeVersion: "1.18", succeed: true, exp: "cis-1.5"},
+		{kubeVersion: "1.16", succeed: true, exp: "cis-1.6"},
+		{kubeVersion: "1.17", succeed: true, exp: "cis-1.6"},
+		{kubeVersion: "1.18", succeed: true, exp: "cis-1.6"},
```
lizrice, line 216: We should probably add 1.19 in now since it has just been released (assume it is also covered by cis-1.6)

cfg/config.yaml
```diff
@@ -194,9 +194,9 @@ version_mapping:
   "1.13": "cis-1.4"
   "1.14": "cis-1.4"
   "1.15": "cis-1.5"
-  "1.16": "cis-1.5"
-  "1.17": "cis-1.5"
-  "1.18": "cis-1.5"
+  "1.16": "cis-1.6"
+  "1.17": "cis-1.6"
+  "1.18": "cis-1.6"
```
lizrice, line 199: I think we should also add in 1.19 since that has just been released

lizrice (approved):
LGTM! we should also make sure it will work on Kubernetes 1.19 which was just released recently, so please make those changes and then let's get this shipped!

lizrice (approved):
lgtm!

PR #689, opened 2020-09-02
Updating section id 4.6

dyapas:
- id: 4.6
        text: "Verify the scheduler pod specification file ownership set by OpenShift"
        audit: "stat -c %u:%g /etc/origin/node/pods/controller.yaml" -- (lower case u and g ) it returns the uID and gID in numeric i.e 0:0 not root:root.
it supposed to be Uppercase: audit: "stat -c %U:%G /etc/origin/node/pods/controller.yaml"


lizrice (approved):
lgtm, thank you!

Issue #671, opened 2020-08-10
EKS "exec format error"


simar7:
_Originally posted by @manjo-git in https://github.com/aquasecurity/kube-bench/pull/628#issuecomment-671417026_

I have EKS deployed with m6.large (ARM64), and I am using  image: aquasec/kube-bench:latest in my jobs-eks.yaml and I get an "exec format error". My jobs-eks.yaml can be found here: https://pastebin.ubuntu.com/p/4MvfX5XHBh/

$ kubectl get pods --all-namespaces
NAMESPACE     NAME                       READY   STATUS              RESTARTS   AGE
default       kube-bench-t62lc           0/1     ContainerCreating   0          2s
default       kube-bench-zwz4r           0/1     Error               0          7s
kube-system   aws-node-fxhxv             1/1     Running             0          6m32s
kube-system   aws-node-mgr89             1/1     Running             0          6m32s
kube-system   aws-node-q7fsq             1/1     Running             0          6m31s
kube-system   coredns-59dd559b88-csjll   1/1     Running             0          18m
kube-system   coredns-59dd559b88-x6t9s   1/1     Running             0          18m
kube-system   kube-proxy-6b958           1/1     Running             0          6m32s
kube-system   kube-proxy-p22qx           1/1     Running             0          6m32s
kube-system   kube-proxy-xx7tl           1/1     Running             0          6m31s

$ kubectl logs kube-bench-zwz4r
standard_init_linux.go:211: exec user process caused "exec format error"

Can you please point me in the right direction as to how to use the public kube-bench image for ARM64 EKS ? The instructions https://github.com/aquasecurity/kube-bench#running-in-an-eks-cluster seems to work only for AMD64 because AWS has command lines tools only on AMD64 and not on ARM64.

lizrice:
Aha, is this on Graviton2 machines in AWS?

simar7:
> Aha, is this on Graviton2 machines in AWS?

hi @manjo-git would you know?

manjo-git:
Yes this is on Graviton2 system in AWS.

manjo-git:
Hello, is there any update on this one? I would like to give it a go on the graviton2 that just went GA on AWS today. https://aws.amazon.com/blogs/containers/eks-on-graviton-generally-available/

lizrice:
[...]

PR #690, opened 2020-09-03
Multi-arch build 

lizrice:
Addresses #671 by building a multi-arch image for arm64 as well as amd64

Also tidied up the makefile a bit, including: 
 - removed duplicate run of unit tests
 - added some dependencies for kind-run tests to make sure it's using an up-to-date image

Note that in the Dockerfile for some reason `go install` created an executable with a different name (not `kube-bench`) on amd64. I've worked around this by using `go build` instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,13 +16,18 @@ before_install:
 
 script:
   - yamllint -c ./.yamllint.yaml .
-  - GO111MODULE=on go test ./...
+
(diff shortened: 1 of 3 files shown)


yoavrotems (approved):
Anything that seems for a second problematic you explained in your comment (like removing the `make tests`) so LGTM

PR #691, opened 2020-09-03
docs: more clarification on output states

### LINKED ISSUE #595: Need help understanding WARN state and scoring mechanism
Hey !

I'm a bit confused about the [WARN] state...  

Can someone explain the difference between "scored" and "not scored" ?

I also don't get how it influences the state ? **Scored** + **unable to test** creates what in the end ??  
And the other way around is not clear either. The worst case that blows each theories that I've got so far is the **warn** + **scrored**. This makes no sense what so ever ! :-p  

Thanks in advanced

lizrice:
Resolves #595

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,16 +284,17 @@ kubectl apply -f job-gke.yaml
 
 ## Output
 
-There are three output states:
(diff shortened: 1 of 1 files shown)


yoavrotems:
LGTM

yoavrotems (approved):

PR #700, opened 2020-09-10
Note about OpenShift OCP 4.*

lizrice:
- Add note about why we don't support OCP 4.*
- Move GKE & OpenShift sub-sections next to EKS and AKS
- Minor corrections


yoavrotems (approved):
LGTM

Issue #701, opened 2020-09-16
When missing a component not running


yoavrotems:
When running kube-bench and missing a component such as kubelet not running.
Missing the component could cause problem to individual tests but not to the entire running of kube-bench since its not necessary for running but for testing.
![image](https://user-images.githubusercontent.com/12915002/93306134-286ed380-f808-11ea-930e-6a3bb7bfb626.png)

PR #702, opened 2020-09-16
Not exiting when executable not found

yoavrotems:
Regrading https://github.com/aquasecurity/kube-bench/issues/701 where kube bench is crushing when not finding components

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -91,7 +91,7 @@ func runChecks(nodetype check.NodeType, testYamlFile string) {
 
 	// Checks that the executables we need for the section are running.
 	if err != nil {
-		exitWithError(fmt.Errorf("failed to get a set of executables needed for tests: %v", err))
+		glog.V(1).Info(fmt.Sprintf("failed to get a set of executables needed for tests: %v", err))
 	}
 
 	confmap := getFiles(typeConf, "config")


cmd/common.go
```diff
@@ -91,7 +91,7 @@ func runChecks(nodetype check.NodeType, testYamlFile string) {
 
 	// Checks that the executables we need for the section are running.
 	if err != nil {
-		exitWithError(fmt.Errorf("failed to get a set of executables needed for tests: %v", err))
+		glog.V(1).Info(fmt.Sprintf("failed to get a set of executables needed for tests: %v", err))
```
lizrice, line 94: Can we still just continue with running the tests in this section? Or should we be returning from this function with this error? I guess we should be returning [FAIL] or [WARN] for each test in this section

lizrice (changes requested):
It's only one line change, but it could have huge consequences! This would be a good example where you can help the reviewer by providing the output before and after the change, using an example where a section fails

cmd/common.go
yoavrotems: Its before running anything, its changing the behavior from not running and run and fail/warn on the relate tests 
This is the output before the change: 
image
After the change we can see that now we can run and get cis output results:
image

lizrice (approved):
OK, I'm convinced by seeing the output change!

PR #703, opened 2020-09-16
Set actualResult

yoavrotems:
actual Result is used later on to get actual value and the --include-test-output values but it never got set so its always empty.

### DIFF
--- a/check/test.go
+++ b/check/test.go
@@ -169,6 +169,7 @@ func (t testItem) execute(s string) *testOutput {
 		}
 	}
 
+	result.actualResult = s
 	return result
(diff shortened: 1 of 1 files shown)


yoavrotems:
Didn't set this value in here https://github.com/aquasecurity/kube-bench/blob/master/check/test.go#L175 where its seems fit because there the output is separated to line and not including the whole output

danielpacak (approved):
LGTM! I don't know how urgent is it, but to practice TDD it would be nice to add assert on the `actualResult` property in the TestTestExecute unit test.

Issue #699, opened 2020-09-09
CIS Benchmark Tests 4.1.3 and 4.1.4 provide dubious answers


distortedsignal:
## What I did (in a nutshell)
```sh
> sudo rpm -i https://github.com/aquasecurity/kube-bench/releases/download/v0.3.1/kube-bench_0.3.1_linux_amd64.rpm
> kube-bench --help
--- kube-bench help redacted ---
> cp -R /etc/kube-bench/cfg .
> kube-bench node --version 1.16 % the cluster is actually version 1.18, but that's fine
...
[PASS] 4.1.2 Ensure that the kubelet service file ownership is set to root:root (Scored)
[FAIL] 4.1.3 Ensure that the proxy kubeconfig file permissions are set to 644 or more restrictive (Scored)
[FAIL] 4.1.4 Ensure that the proxy kubeconfig file ownership is set to root:root (Scored)
[PASS] 4.1.5 Ensure that the kubelet.conf file permissions are set to 644 or more restrictive (Scored)
...
== Remediations ==
4.1.3 Run the below command (based on the file location on your system) on the each worker node.
For example,
chmod 644 /etc/kubernetes/proxy.conf

4.1.4 Run the below command (based on the file location on your system) on the each worker node.
For example, chown root:root /etc/kubernetes/proxy.conf
...
> ps -ef | grep kube-proxy | grep -v grep
root     23174 23158  0 09:58 ?        00:00:03 /usr/local/bin/kube-proxy --config=/var/lib/kube-proxy/config.conf --hostname-override=mip-bd-vm217.mip.storage.hpecorp.net
```

## What I expected to happen
```sh
> sudo rpm -i https://github.com/aquasecurity/kube-bench/releases/download/v0.3.1/kube-bench_0.3.1_linux_amd64.rpm
> kube-bench --help
--- kube-bench help redacted ---
> cp -R /etc/kube-bench/cfg .
> kube-bench node --version 1.16 % the cluster is actually version 1.18, but that's fine
...
[PASS] 4.1.2 Ensure that the kubelet service file ownership is set to root:root (Scored)
[PASS] 4.1.3 Ensure that the proxy kubeconfig file permissions are set to 644 or more restrictive (Scored)
[PASS] 4.1.4 Ensure that the proxy kubeconfig file ownership is set to root:root (Scored)
[PASS] 4.1.5 Ensure that the kubelet.conf file permissions are set to 644 or more restrictive (Scored)
...
[...]

PR #708, opened 2020-09-24
Improve Proxykubeconfig tests

distortedsignal:
A while ago I opened #699 and got a lot of great help from the Aqua Security team. I was hoping I would be able to contribute my changes back to the source.

## What this PR does

Previously, if the `proxykubeconfig` was loaded into the kube-proxy using a configmap, cis-1.5 and cis-1.6 tests 4.1.3 and 4.1.4 would be falsely failed, and cis-1.3 and cis-1.4 tests 2.2.5 and 2.2.6 tests would also be falsely failed. With these changes, I believe that all tests should operate as intended by the CIS team to provide better results to kube-bench users.

Please note that cis-1.3/4 test 2.2.5 is very similar to cis-1.5/6 test 4.1.3 and that cis-1.3/4 test 2.2.6 is very similar to cis-1.5/6 test 4.1.4. The rest of this PR is going to be breaking these tests down by these dichotomies. 

### Changes in 2.2.5/4.1.3

This is by far the easier change, and was very neatly laid out by @lizrice in #699 in [this comment](https://github.com/aquasecurity/kube-bench/issues/699#issuecomment-693502903). So instead of saying "the `--kubeconfig` flag must be passed in and the file that it points to must have these permissions," the code now says "if the `--kubecconfig` flag is passed in, the file must have these permissions, if the flag is not passed in, pass the test." I believe that this is more in the spirit of the CIS document (as was presented in the GitHub issue).

### Changes in 2.2.6/4.1.4

This was an uglier change that became really nice after more thought. If the file does not exist, this test should pass (similar to 2.2.5/4.1.3). The way this test is currently written, if the file is not owned by `root:root`, then the test fails, even if the file does not exist. So if the variable that notes where the file should be doesn't exist, then I say that the test is passed.

## The outcome

```sh
[root@k8s-node-host ~]# cat /etc/kube-bench/cfg/cis-1.5/node.yaml | tail -n 445 | head -n 33
      - id: 4.1.3
        text: "Ensure that the proxy kubeconfig file permissions are set to 644 or more restrictive (Scored)"
        audit: '/bin/sh -c ''if test -e $proxykubeconfig; then stat -c permissions=%a $proxykubeconfig; fi'' '
        tests:
          bin_op: or
          test_items:
            - flag: "permissions"
              set: true
              compare:
                op: bitmask
                value: "644"
            - flag: "permissions"
              set: false
        remediation: |
          Run the below command (based on the file location on your system) on the each worker node.
          For example,
          chmod 644 $proxykubeconfig
        scored: true

      - id: 4.1.4
        text: "Ensure that the proxy kubeconfig file ownership is set to root:root (Scored)"
        audit: '/bin/sh -c ''if test -n "$proxykubeconfig"; then stat -c %U:%G $proxykubeconfig;  fi'' '
        tests:
          bin_op: or
          test_items:
            - flag: root:root
              set: true
            - flag: "proxykubeconfig"
              set: false
        remediation: |
          Run the below command (based on the file location on your system) on the each worker node.
          For example, chown root:root $proxykubeconfig
        scored: true
[root@k8s-node-host ~]# ps -ef | grep kube-proxy | grep -v grep
root     35222 35199  0 16:30 ?        00:00:01 /usr/local/bin/kube-proxy --config=/var/lib/kube-proxy/config.conf --hostname-override=k8s-node-host.company.net
[root@k8s-node-host ~]# kube-bench node --version="1.17"
[INFO] 4 Worker Node Security Configuration
...
[PASS] 4.1.2 Ensure that the kubelet service file ownership is set to root:root (Scored)
[PASS] 4.1.3 Ensure that the proxy kubeconfig file permissions are set to 644 or more restrictive (Scored)
[PASS] 4.1.4 Ensure that the proxy kubeconfig file ownership is set to root:root (Scored)
[PASS] 4.1.5 Ensure that the kubelet.conf file permissions are set to 644 or more restrictive (Scored)
...
[root@k8s-node-host ~]#
```

And that, I think, is pretty cool.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,26 +17,23 @@ before_install:
     if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then
       echo "$DOCKER_PASSWORD" | docker login --username "$DOCKER_USERNAME" --password-stdin
     fi
-
 script:
   - yamllint -c ./.yamllint.yaml .
+
+  # Run unit and integration tests
+  - make tests
+  - make integration-tests
 
   # Build a local container image to test that the install sub-command works
   - IMAGE_NAME=kube-bench make build-docker
   - docker run -v `pwd`:/host kube-bench install
   - test -d cfg
   - test -f kube-bench
-
-  # Run unit and integration tests
-  - make tests
-  - make integration-tests
-  
   # Build and push the multi-arch Docker image
   - |
     if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then
       make docker
     fi
-
 after_success:
   - bash <(curl -s https://codecov.io/bash)
 deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
     if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then
       echo "$DOCKER_PASSWORD" | docker login --username "$DOCKER_USERNAME" --password-stdin
     fi
-    
+
 script:
   - yamllint -c ./.yamllint.yaml .
 
@@ -35,7 +35,7 @@ script:
     if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then
       make docker
     fi
-    
+
 after_success:
   - bash <(curl -s https://codecov.io/bash)
 deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ before_install:
     if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then
       echo "$DOCKER_PASSWORD" | docker login --username "$DOCKER_USERNAME" --password-stdin
     fi
+    
 script:
   - yamllint -c ./.yamllint.yaml .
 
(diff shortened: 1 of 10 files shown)


distortedsignal:
So... the tests failed, but I don't think this change caused them to fail.

Screen Shot 2020-09-25 at 3 21 22 PM

If I did something to your docker password... I'm _REALLY_ sorry.

distortedsignal:
Hey, I would really like to get this merged before the next version comes out - how likely/possible is that?

lizrice (changes requested):
Looks good to me. Please can you resolve the conflict (it's because we're retiring the old cis-1.3 and 1.4 files) and then I think we'll be good to go

lizrice:
Sorry about the docker password issue by the way @distortedsignal, that was our fault!

yoavrotems:
Hey, we fixed the issue with docker password but its seems to fail because you didn't update the .data files in https://github.com/aquasecurity/kube-bench/tree/master/integration/testdata

distortedsignal:
@yoavrotems Yeah, I saw that some other tests were failing - going to work on those now. That's why this is a "draft" as opposed to an actual PR.

distortedsignal:
@yoavrotems I might need more help here than I initially thought. I tried to manually update those files, but it looks like my changes didn't take effect for one reason or another. Would you be willing to take a look at why they're failing?

distortedsignal:
I'm working on WSL for this, so it's quite challenging to test this locally. Would someone mind advising me on the fastest way forward?

distortedsignal:
This isn't worth my time - let me know when you figure out how to get those files populated into containers @yoavrotems.

lizrice:
I'm cutting v0.4.0 now to make sure we have a release that includes CIS 1.6 test files, but we can cut another release 0.4.1 when this is ready

distortedsignal:
Thanks @yoavrotems for helping me get everything sorted. Sorry it took so long to get this into a merge-able state.

yoavrotems (approved):
LGTM! 
Thanks for the contribution :)

lizrice (approved):
LGTM, thanks for pushing this over the finish line @distortedsignal!

PR #709, opened 2020-09-28
Remove weak ciphers from validation

pjbgf:
The ciphers below are considered weak and it would be good to remove from the current recommendation:

`TLS_RSA_WITH_AES_256_GCM_SHA384` and `TLS_RSA_WITH_AES_128_GCM_SHA256`

Fixes https://github.com/aquasecurity/kube-bench/issues/710

This is also aligned with the Kubernetes Security audit, which also does not recommend the use of the above ciphers.
https://robotattack.org/ 
https://github.com/kubernetes/community/blob/master/wg-security-audit/findings/Kubernetes%20Final%20Report.pdf

Link to CIS benchmark ticket: 
https://workbench.cisecurity.org/benchmarks/6083/tickets/11462


lizrice:
i completely see the value of tightening up the checks, but we do want the test files to reflect what’s actually in the CIS benchmark as closely as possible. The test files are named to reflect different versions of the CIS benchmarks. Therefore we won’t be able to accept these changes as-is, and I’ll close this PR for now. I am going to move the corresponding issue to a discussion though so we can think about a different way to address this idea.

PR #706, opened 2020-09-17
upgrade to go version 1.15

neha-viswanathan:
Upgrade golang base image to 1.15

PR #713, opened 2020-10-01
Test Travis build condition

yoavrotems:
Currently when PR from a fork its fails travis build because of security issues with the environment variables.
Used the changes made in https://github.com/aquasecurity/kube-bench/pull/706 
and just updated the travis file.


lizrice (approved):
let's try it!

PR #693, opened 2020-09-07
retire cis 1.3 and 1.4

neha-viswanathan:
closes #679 

This MR removes unsupported K8s versions - cis 1.3 and 1.4 tests

PR #727, opened 2020-10-07
Remove Unnecessary CIS-1.3 and -1.4 data files.

distortedsignal:
Let's drop that code coverage percentage!

Because the CIS-1.3 and -1.4 tests are no longer run by the `kube-bench` tool, the repository no longer needs data files related to the output of those tests. I wanted to make sure that future maintainers didn't make changes to files that don't need to be changed. Also because I feel like this:

4ho9aa


yoavrotems (approved):
Seems like something that should of happened here https://github.com/aquasecurity/kube-bench/pull/693.
Good job noticing that. LGTM

distortedsignal:
I just want to make sure that your codebase is as lean and easy to work on as possible. Seems like a cool project!

Issue #723, opened 2020-10-05
Add —exit-code 


lizrice:
At the moment, if kube-bench runs successfully, it exits with a zero exit code even if there are failed tests. Let’s add a `—exit-code` option, much like already exists for Trivy. If `—exit-code` is specified, and any of the tests result in FAIL, exit the kube-bench executable with the value specified by `exit-code`

See discussion #566

malathit:
@lizrice I am a developer with some experience on kubernetes. Recently, I saw your lecture in kubecon 2020 and it was wonderful. I would like to work on this and get this done in a week or two max. Can you please assign this issue so that I can work on this if resolving this is not urgent?

Wicked7000:
As there wasn't an update on this Issue for a couple of days I decided to go ahead and make the set of changes, please let me know if they are acceptable or if anything else is needed. (Or if someone else is going to pick up the changes)
Thanks!

PR #734, opened 2020-10-10
Add exit-code parameter for when checks have failed

Wicked7000:
This is a change that addresses the issue #723

**Behaviour:**
- If exit-code parameter is unset it defaults to 0, meaning regardless of pass or failure it always returns exit code 0. (Assuming kube-bench made it to the end of execution without error)
- If exit-code is specified and is non-zero then if any control collection has more than 0 failures it will exit with the supplied exit code.

Please let me know if any additional changes are needed or if there is any change in functionality needed. I've tested locally and it all seems to work correctly.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -337,16 +337,12 @@ func isThisNodeRunning(nodeType check.NodeType) bool {
 }
 
 func exitCodeSelection(controlsCollection []*check.Controls) int {
-	checksFailed := false
 	for _, control := range controlsCollection {
-		if control.Fail > 0 {
-			checksFailed = true
+		if control.Fail > 0  {
+			return exitCode
 		}
 	}
 
-	if checksFailed && exitCode != 0 {
-		return exitCode
-	}
 	return 0
 }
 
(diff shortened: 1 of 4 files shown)


cmd/common.go
```diff
@@ -336,6 +336,20 @@ func isThisNodeRunning(nodeType check.NodeType) bool {
...
+		if control.Fail > 0 {
+			checksFailed = true
+		}
+	}
+
+	if checksFailed && exitCode != 0 {
+		return exitCode
+	}
+	return 0
+}
```
rmsy, line 351: You might simplify this like so:
```suggestion
func exitCodeSelection(controlsCollection []*check.Controls) int {
	for _, control := range controlsCollection {
		if control.Fail > 0 {
			return exitCode
		}
	}

	return 0
}
```

Wicked7000: Yup totally agree, I've changed it to that now

guusvw (approved):

Wicked7000:
@rmsy @guusvw Is there any additional changes that are needed or is this okay to be merged now?

yoavrotems (approved):
LGTM thanks for contributing!

Issue #649, opened 2020-07-28
Support type: "skip" for a group of checks


lizrice:
At the moment the user can skip an individual check by setting it's `type` field to `"skip"`. We could extend this support to the group level so that it's easy to skip a whole group of tests.

Wicked7000:
I've opened a PR for this please have a look when you get a chance and let me know if anything needs changing, many thanks!

PR #736, opened 2020-10-12
Allow for skip to be defined on a group-level skipping all checks inside

Wicked7000:
This fixes issue #649 allowing groups to be defined to be skipped. this is done by setting 'Skip' on the group to true. When checks are run this will then set the state of any checks within that group to info (the check is still technically run but just overridden to be INFO as I wasn't sure on what the desired behaviour here was, we could just skip the check entirely and not run and just set it to INFO).

Please let me know if there are any additional changes you want done to this!

### DIFF
--- a/check/check.go
+++ b/check/check.go
@@ -39,6 +39,9 @@ const (
 	// INFO informational message
 	INFO State = "INFO"
 
+	// SKIP for when a check should be skipped.
+	SKIP = "skip"
+
 	// MASTER a master node
 	MASTER NodeType = "master"
 	// NODE a node
@@ -111,7 +114,7 @@ func (c *Check) run() State {
 	}
 
 	// If check type is skip, force result to INFO
-	if c.Type == "skip" {
+	if c.Type == SKIP {
 		c.Reason = "Test marked as skip"
 		c.State = INFO
 		return c.State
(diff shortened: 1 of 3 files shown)


yoavrotems:
Hey first of all thanks for contributing, secondly your doubt was right If a test is skip we don't want to run it at all. 
And I think instead of adding new code the handle skip for group and change the state to INFO as you did maybe we should just inherit it the the group tests? Like if we have a group marked as skip then we could add the skip to every test inside the group so we will still handle it correctly and take care of the test as needed for example https://github.com/aquasecurity/kube-bench/blob/master/check/check.go#L113 where we add the right reason for the test state.
WDYT ?

Wicked7000:
> Hey first of all thanks for contributing, secondly your doubt was right If a test is skip we don't want to run it at all.
> And I think instead of adding new code the handle skip for group and change the state to INFO as you did maybe we should just inherit it into the group tests? Like if we have a group marked as skip then we could add the skip to every test inside the group so we will still handle it correctly and take care of each test as needed for example https://github.com/aquasecurity/kube-bench/blob/master/check/check.go#L113 where we add the right reason for the test state.
> WDYT ?

Yeah apologies that makes sense, I've changed the code to just switch over the type to skip and adjusted the tests for that. (it uses a normal runner rather than mocking it otherwise I don't think it would really be testing anything, but I don't think that should be a problem as its not running the actual test code anyways)

Let me know if you think anything else needs to be changed!

yoavrotems (approved):
LGTM, really good job and thanks for contributing 👍

PR #738, opened 2020-10-12
Customize kubeconfig locations for kube-scheduler and kube-controller-…

bjrara:
**Why do we need it?**
Avoid hard code kubeconfig locations for kube-scheduler and kube-controller-manager in the check. Introduce schedulerkubeconfig and controllermanagerkubeconfig variable to commands so that users can customize kubeconfig files in their clusters.


yoavrotems (approved):
LGTM thank you for contributing !

PR #739, opened 2020-10-12
Fix tls-cipher-suites check for kube-apiserver

bjrara:
**Why do we need it?**
According to CIS Benchmark, item 1.2.35 allows kube-apiserver to configure with a subset of the given cipher suites. 
```
Edit the API server pod specification file /etc/kubernetes/manifests/kube-apiserver.yaml
on the master node and set the below parameter as follows, or to a subset of these values.
```
However, using the current configuration, it checks whether this whole set is the same set or is a subset of the cipher suites configured in the kube-apiserver which violates the description. This PR changes `has` op to `valid_elements` to correct the check logic.

### DIFF
--- a/cfg/cis-1.5/master.yaml
+++ b/cfg/cis-1.5/master.yaml
@@ -908,7 +908,7 @@ groups:
           test_items:
             - flag: "--tls-cipher-suites"
               compare:
-                op: has
+                op: valid_elements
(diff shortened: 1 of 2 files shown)


yoavrotems (approved):
Hey good catch thank you :) !

PR #740, opened 2020-10-12
interactive code change request

anandprabhakar0507:


yoavrotems:
Thanks for contributing but we can get all needed information about the program state with -v flag (print logs) so this PR add no value, but you are more then welcome to seek out in [issues ](https://github.com/aquasecurity/kube-bench/issues)and help with something from there :)

PR #742, opened 2020-10-14
Automate check 3.2.1 Ensure that a minimal audit policy is created

bjrara:
**Why do we need it?**
This PR automates the verification of `3.2.1. Ensuring a minimal audit policy is created` by validating `--audit-policy-file` is set to kube-apiserver start command.

If this option is configured, the kube-apiserver will check the validity of the policy file during startup.
Quote from [Auditing](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#audit-policy):
> Note that the rules field must be provided in the audit policy file. A policy with no (0) rules is treated as illegal.

So it's safe to mark the item as PASS as long as it is set.

### DIFF
--- a/cfg/cis-1.5/controlplane.yaml
+++ b/cfg/cis-1.5/controlplane.yaml
@@ -21,7 +21,11 @@ groups:
     checks:
       - id: 3.2.1
         text: "Ensure that a minimal audit policy is created (Scored)"
-        type: "manual"
+        audit: "/bin/ps -ef | grep $apiserverbin | grep -v grep"
+        tests:
+          test_items:
+            - flag: "--audit-policy-file"
+              set: true
         remediation: |
           Create an audit policy file for your cluster.
         scored: true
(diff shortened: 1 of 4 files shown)


bjrara:
This PR is open for discussion. Feel free to add any thoughts.

yoavrotems:
Hey seems good! But why just adding it to version 1.5, why not to all of them?

bjrara:
@yoavrotems From my understanding, this item can be automated. But in CIS 1.6, the description changed to `[WARN] 3.2.1 Ensure that a minimal audit policy is created (Manual)`. So I'm if there's anything beyond my knowledge makes CIS to change the guideline. I added to 1.5 because I'm not sure if it can be a global-wise change. I need more opinions on this.

yoavrotems:
https://github.com/aquasecurity/kube-bench/discussions/664 I think this discussion start to talk about this issue with manual and not scored

bjrara:
@yoavrotems Thanks for sharing the context. I updated `3.2.1 auditing` in both 1.5 and 1.6, and left `scored: false` in 1.6, leaving users to decide whether it should be automated.

yoavrotems (approved):
LGTM !

PR #746, opened 2020-10-15
Create a Heart

rohit-rksaini:
Make a heat using *


yoavrotems:
Thank you for contributing but it is not relate to project in any way.
You are more then welcome to take on of the open issues and help there :)

PR #748, opened 2020-10-18
Optimize images

JeffResc:
This pull request optimizes the images found in the repository using lossless compression. This means that the image quality will not be degraded. This will mean less space taken up by images along with faster loading times when viewing them. The overall file size has been reduced by around 46%.


yoavrotems:
Hey Thanks for this contribution, but it will be unnecessary, you are more then welcome to seek something in out issues and help there! :)

Issue #564, opened 2020-01-13
Add option to ignore check-failures


mrueg:
Some checks in the kube-bench might currently not be feasible for a user and the user wants to ignore them. 
Currently the user can only specify all tests except for those via `--check`.
It would be good if kube-bench could provide a CLI argument, that allows to soft-fail on specified tests, so they still would run but would not alter the return code:

1.1.1.2 and 1.1.1.3 fail, a user could run:
```kube-bench --expect-fail "1.1.1.2,1.1.1.3"``` and still the return code would be 0 (if all other checks pass)

lizrice:
Another option is to modify the test YAML, would that suit your needs?

mrueg:
I'd like to avoid that, maintaining numbers for the command seems to be easier than patching the yaml from a user perspective (especially with a kube-bench container using a readonly rootfs, although I have to admit this can of course be changed at buildtime).

lizrice:
Another idea: `--skip` which has a comma separated list of tests that should be treated as if they were marked with "skip" - wdyt?

mrueg:
skip works for the use case as well. soft-fail would allow to make that list smaller and smaller over time.

lizrice:
I think we should go with `—skip` because it’s consistent with the terminology we use inside the YAML file, and it would have the same effect - kube-bench would just not run a test if it’s marked as `skip` either in YAML or with this parameter

illyaMs:
Trying to figure out the draft implementation. Some questions so far (would be nice to have feedback/answers from maintainers):

1. Do I get it right that a proper place to introduce this new logic is the [filtering predicate](https://github.com/aquasecurity/kube-bench/blob/35cf28c140bdb6f84c3a29749780a08594300f69/cmd/common.go#L31)?
1. I see that having both `--group` and `--check` options at the same time is [explicitly prohibited](https://github.com/aquasecurity/kube-bench/blob/35cf28c140bdb6f84c3a29749780a08594300f69/cmd/common.go#L33). Should it be the same for this new `--skip` option? Or it is ok to have for example both `--skip` and `--group` options at the same time?
1. Should we have separate options like `--skipGroup` and `--skipCheck`, or having just `--skip` to support both filterings is enough?
1. Should `--skip` option take precedence over the `skip` field in test's yaml definition?

Thanks in advance.

iahmad-khan:
Hi,

[...]

PR #751, opened 2020-10-21
Allow for groups and checks to be skipped via command line parameter

Wicked7000:
This adds the feature specified in the PR #564 

Currently this is how the workflow works:
- User can add --skip followed by any number of IDS 
- For each id before running checks it will see if it belongs to the current group or check.
- If it belongs to the current group all child checks are skipped, if it belongs to the current check only that current check is skipped.
- If skip is specified in the yaml or the command line parameter it will be skipped.

Although the original PR did have a comment specifying that 1.2.* for example could be good I wasn't sure if that was wanted so if any maintainer could let me know what they think of the current implementation and I can go from there, thanks!

### DIFF
--- a/check/controls.go
+++ b/check/controls.go
@@ -75,7 +75,7 @@ func NewControls(t NodeType, in []byte) (*Controls, error) {
 }
 
 // RunChecks runs the checks with the given Runner. Only checks for which the filter Predicate returns `true` will run.
-func (controls *Controls) RunChecks(runner Runner, filter Predicate) Summary {
+func (controls *Controls) RunChecks(runner Runner, filter Predicate, skipIdMap map[string]bool) Summary {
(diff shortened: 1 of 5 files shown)


yoavrotems (approved):
Hey I really like what you done, its really simple and straight forward, I think that its better the have it without the *, and just as you did, 1.2 then it get 1.2 group and if you did, 1.2.1 then it will be 1.2.1 check,
Since each group suppose to follow the same logic (CIS chapters) then It wouldn't make sense to add *, the only use for it will be 1.2.* and it could lead to a wrong use of 1.*.1 for example where un logically all 1.1.1 1.2.1 1.3.1 tests will be skipped. 
Long story short its seems good and thanks for contributing! 👍

Issue #667, opened 2020-08-08
Support detecting etcd arguments via env vars


rifelpet:
Kops uses environment variables to configure etcd arguments like cert-file and key-file. See all of the `ETCD_*` env vars documented [here](https://etcd.io/docs/v3.4.0/op-guide/configuration/) as well as this example:

https://github.com/kopeio/etcd-manager/blob/26ca2ad568e9819f04cf00f43d79bd76c5f9dfb3/pkg/etcd/etcdprocess.go#L229-L241

Currently kube-bench reports that these arguments are not set because only process arguments are checked:

https://github.com/aquasecurity/kube-bench/blob/a6161aa8684818157d583e1e64b56d4c6b51046d/cfg/cis-1.5/etcd.yaml#L13-L20

It would be great to also check the etcd process' environment variables in determining whether they're set.

ankitsrao:
I agree with @rifelpet . Maybe this is the reason why, for the etcd policies, the status remains FAIL, even though we follow the remediation procedure.
Anyone else facing the same issue?

lizrice:
I've raised a ticket about this on the CIS benchmark too because the recommendations should really reflect this as well. 

Getting the environment variables for another process is possible via the `/proc` directory on the host but this is not going to be a trivial change to make. (Unless there is a way to query etcd's configuration over an API?)

In the mean time you may prefer to [skip](https://github.com/aquasecurity/kube-bench#omitting-checks) these tests

PR #755, opened 2020-10-23
Allow for environment variables to be checked in tests

Wicked7000:
This addresses the issue #667 and allows for environment variables to be checked from tests it currently works as below:

**Overall behaviour:**
- Attempt to use audit to find the flag and pass the test
- Attempt to use config to find the flag and pass the test (assuming the flag was not found)
- Attempt to use environment variables to find the flag and pass the test (assuming the flag was not found)
  - This is only run if 'Env' is specified as part of the test item meaning that the data its checking can also be set within an environment variable.
  - It gets the environment variables for the process by running `AuditEnv` from the yaml (or the default, more on this below)
  - If the environment variable is empty it counts as not found, otherwise the value and match state are passed back to the execute code for the test to run it on the operators.

The way it uses the environment variable is the same way it would use the value for an argument (so all operators work such as noteq and eq... etc) some of the expected text has changed to make a little more sense too. Keep in mind the way this is implemented means that if env and flag are specified on the same test item they are equivalent to an 'OR' (meaning pass if the flag is found or the env is found, along with matching to the operator) 

I've currently not got around to adding tests (but will likely be getting to them in the next day or two). However I wanted to get this PR out there as I had some questions around the implementation namely:

**Questions for maintainers:**
- I think the overall default command (example below) that gets the environment variables should work on all linux distros (TR and CAT is part of GNU Coreutils) and as the proc directory itself is handled by the linux kernel I assume this should work fine. However if anyone has concerns or otherwise please let me know. `cat "/proc/$(/bin/ps -C etcd -o pid= | tr -d ' ')/environ" | tr '\0' '\n'`
- Do we need to remove quotes from around env variables. I have checked this on KDE Neon and it seems that quotes are removed when calling proc anyways so I don't think this would be an assume but would like to hear opinions on it.
- Currently a flag passing means environment check will not be run (as it assumes argument > env). I think this is fine as generally this is the pattern (such as with etcd) but there may be services in the future that don't abide by such behaviour.
- Currently AuditEnv does not need to be specified in the yaml files due to the fact it is likely going to be exactly the same for every test that wants to use it (Creating a large amount of likely un-wanted duplication). The approach that was used is to create an 'default' AuditEnv. This is only created in situations where there was a single bin file subbed into the Audit command otherwise an explict AuditEnv is required.

I have tested most of the checks against a running locally instance of etcd and it all seems to work well. 
Additional note: I'll also squash commits once tests are written and I can get some insight on the above information to know if there is any additional changes that might be needed.


check/check.go
```diff
@@ -184,6 +186,11 @@ func (c *Check) run() State {
 }
 
 func (c *Check) runAuditCommands() (lastCommand string, err error) {
+	// Always run auditEnvOutput if needed
+	if c.AuditEnv != "" {
+		c.AuditEnvOutput, err = runAudit(c.AuditEnv)
```
lizrice, line 191: Should handle this error if non-nil

check/check.go
```diff
@@ -212,6 +219,10 @@ func (c *Check) execute() (finalOutput *testOutput, err error) {
 		if !result.flagFound {
 			t.isConfigSetting = true
 			result = *(t.execute(c.AuditConfigOutput))
+			if !result.flagFound && t.Env != "" {
+				t.isConfigSetting = false
```
lizrice, line 223: Currently if `isConfigSetting` is false it means we get it from running `c.auditOutput`, and if it's true it means we get it from running `c.auditConfigOutput`. I think we need to change this from a boolean to an enum so that it can take three options: `audit`, `auditConfig` or `auditEnv`. Maybe we change the first one to be `auditCommand` for even more clarity. 

Then this enum is what you can use in `t.findValue()` (see other comment)

check/test.go
```diff
@@ -176,7 +200,17 @@ func (t testItem) execute(s string) *testOutput {
 func (t testItem) evaluate(s string) *testOutput {
 	result := &testOutput{}
 
-	match, value, err := t.findValue(s)
+	match, value, err := t.findFlag(s)
+	if err != nil || !match {
```
lizrice, line 204: I think we can avoid changes (or at least reduce) changes to this function by having `t.findValue(s)` work for all three types (currently it does two, so let's add Env into it if possible). 

So I think this code wouldn't need to change, but just in case it does - if this returns there's an error, let's handle it. That's a different case from not finding a match.

check/test.go
```diff
@@ -83,20 +85,42 @@ func (t testItem) flagValue() string {
...
 
+	if t.Env != "" {
+		return fmt.Sprintf("%s or %s", t.Flag, t.Env)
+	}
+
 	return t.Flag
 }
 
-func (t testItem) findValue(s string) (match bool, value string, err error) {
+func (t testItem) findEnv(s string) (match bool, value string) {
```
lizrice, line 95: I'd prefer we use the same pattern for finding values from Env as we currently do for Audit (command line parameters) and AuditConfig (JSON settings). Much as we currently have `pathTestItem` and `flagTestItem` we would have a third option `envTestItem`.  We'd still have a function `t.findValue(s string)` and there would be fewer changes (possibly no changes) in `t.evaluate()`

check/test.go
```diff
@@ -194,8 +228,8 @@ func (t testItem) evaluate(s string) *testOutput {
 		result.testResult = !match
 	}
 
-	result.flagFound = match
 	glog.V(3).Info(fmt.Sprintf("flagFound %v", result.flagFound))
```
lizrice, line 231: I might be mistaken but I think `flagFound` means "we found a value defined for this setting", however it was defined (audit command or config setting). I think finding it from the environment would be an alternative way of setting this flag, I don't think it needs its own `envFound` flag

lizrice (changes requested):
Thanks for this! I have a few comments in the code but I'm happy with the general approach. 

To answer your questions: 

> I think the overall default command (example below) that gets the environment variables should work on all linux distros (TR and CAT is part of GNU Coreutils) and as the proc directory itself is handled by the linux kernel I assume this should work fine. However if anyone has concerns or otherwise please let me know. cat "/proc/$(/bin/ps -C etcd -o pid= | tr -d ' ')/environ" | tr '\0' '\n'

I haven't thought this through very carefully, but my gut feel is that we may find some issues here somewhere! We have had issues before with running `ps` which iirc were overcome by running the version of ps from the kube-bench container (ie.e from alpine). 

I am wondering if we need a configuration setting to allow users to turn off the env changing altogether so that they can easily bypass any issues. That way, if they are not using environment variables for these settings anyway, they can sidestep the attempt to run these commands altogether.  

> Do we need to remove quotes from around env variables. I have checked this on KDE Neon and it seems that quotes are removed when calling proc anyways so I don't think this would be an assume but would like to hear opinions on it.

Do you mean in the YAML definition? I don't think so, it just gets interpreted as a string, and given the env variable names don't contain any special characters I think they get loaded just the same with or without the double quotes. 

> Currently a flag passing means environment check will not be run (as it assumes argument > env). I think this is fine as generally this is the pattern (such as with etcd) but there may be services in the future that don't abide by such behaviour.

Let's document this somewhere

> Currently AuditEnv does not need to be specified in the yaml files due to the fact it is likely going to be exactly the same for every test that wants to use it (Creating a large amount of likely un-wanted duplication). The approach that was used is to create an 'default' AuditEnv. This is only created in situations where there was a single bin file subbed into the Audit command otherwise an explict AuditEnv is required.

This looks neat to me. Let's document it though.

check/test.go
Wicked7000: I agree with this however I think if that is the definition of flagFound it would likely be better to be call it 'found' or something similar (as now its not necessarily finding a flag)

check/check.go
Wicked7000: ✔️

check/check.go
Wicked7000: ✔️

check/test.go
Wicked7000: ✔️

Wicked7000:
I think I've addressed all the issues mentioned above and will start working on tests now, but just as an overview of what was done. `disableEnvTesting` was added as a check level option to skip the auto-generation of the env command to avoid issues where it may cause checks to fail.

README was updated with details about the env operator. All of the code changes requested above were completed.

cmd/util_test.go
```diff
@@ -394,7 +394,7 @@ func TestMakeSubsitutions(t *testing.T) {
 	}
 	for _, c := range cases {
 		t.Run(c.input, func(t *testing.T) {
-			s := makeSubstitutions(c.input, "bin", c.subst)
+			s, _ := makeSubstitutions(c.input, "bin", c.subst)
```
lizrice, line 397: Would be nice to test that the substitutions returned are as expected

lizrice (commented):
I haven't tried it out yet but from a code read this is looking great! Are you still plannint to add some more tests?

cmd/util_test.go
Wicked7000: Yup, totally agree I'm writing tests at the moment apologise for the slight delay should get them pushed tonight.

Wicked7000:
@lizrice This all seems to be good now, please let me know if you think there would be any good additional tests or if you think something else needs changing!

lizrice (approved):
lgtm, thank you!

PR #756, opened 2020-10-24
Don't use variable in remediation

mozillazg:
These variables are imported via #738 and they should not be used in remediation because of the variables in remediation will not be replaced with actual value.

### DIFF
--- a/cfg/cis-1.5/master.yaml
+++ b/cfg/cis-1.5/master.yaml
@@ -234,7 +234,7 @@ groups:
         remediation: |
           Run the below command (based on the file location on your system) on the master node.
           For example,
-          chmod 644 $schedulerkubeconfig
+          chmod 644 /etc/kubernetes/scheduler.conf
         scored: true
 
       - id: 1.1.16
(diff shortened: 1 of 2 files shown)


bjrara:
@mozillazg Would you mind explaining why they will not be replaced in remediation? Did you use the new config.yaml in the PR to verify? https://github.com/aquasecurity/kube-bench/pull/738/files#diff-ee8f320b6fba120f9fd902cea5c861cfa04980ba75ca380943ce3a17a1f04b67

bjrara:
And also I'd like to understand why it's a **needless** customization. We had real cases that we wanted to use kube-bench to audit our secured clusters, but our kubeconfig files were located in other locations than the default one. Those hard-coded controllermanager and scheduler configurations blocked our validation.

bjrara:
@mozillazg Thanks for the report. The variables should be replaced. The unexpected behavior is caused by the misconfiguration in config.yaml.
Fixed in https://github.com/aquasecurity/kube-bench/pull/758.

mozillazg:
@bjrara Get it.

PR #758, opened 2020-10-28
Fix defaultkubeconfig in config.yaml to resolve variable exposure in remediation

bjrara:
**Why do we need it?**
This PR aims at fixing the bug reported in https://github.com/aquasecurity/kube-bench/pull/756.

**How to reproduce this bug?**
The `defaultkubeconfig` for scheduler and controllermanager don't take effect as they are mistakenly configured as an array.

When no files can be found in the declared `kubeconfig` location options. The substitution should fall back to replacing the variables with the default values, however, as the object type is mismatched, so there's no replacement to take place.
```
scheduler:
    kubeconfig:
      - /etc/kubernetes/scheduler.conf
    defaultkubeconfig: /etc/kubernetes/scheduler.conf
```
The issue is initially introduced in https://github.com/aquasecurity/kube-bench/pull/738.

### DIFF
--- a/cfg/config.yaml
+++ b/cfg/config.yaml
@@ -48,8 +48,7 @@ master:
     defaultconf: /etc/kubernetes/manifests/kube-scheduler.yaml
     kubeconfig:
       - /etc/kubernetes/scheduler.conf
-    defaultkubeconfig:
-      - /etc/kubernetes/scheduler.conf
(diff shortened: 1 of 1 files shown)


bjrara:
cc @mozillazg 
/assign @yoavrotems

yoavrotems (approved):
Nice catch :) thank you, LGTM

PR #759, opened 2020-10-29
Add a total summary and always show all tests.

### LINKED ISSUE #528: Summary should cover all tests
When you run kube-bench without specifying the node type, on a master node it runs the master tests, outputs a summary of those, then runs the node tests and outputs a separate summary of those. It would be better if there was one summary covering all the tests that have been run

czunker:
Whether the total summary is shown can be specified with an option.

Fixes #528

Signed-off-by: Christian Zunker <christian.zunker@codecentric.cloud>

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -148,7 +148,7 @@ func generateDefaultEnvAudit(controls *check.Controls, binSubs []string){
 }
 
 func parseSkipIds(skipIds string) map[string]bool {
-	var skipIdMap =  make(map[string]bool, 0)
+	var skipIdMap = make(map[string]bool, 0)
 	if skipIds != "" {
 		for _, id := range strings.Split(skipIds, ",") {
 			skipIdMap[strings.Trim(id, " ")] = true
@@ -185,7 +185,7 @@ func prettyPrint(r *check.Controls, summary check.Summary) {
 	// Print remediations.
 	if !noRemediations {
 		if summary.Fail > 0 || summary.Warn > 0 {
-			colors[check.WARN].Printf("== Remediations ==\n")
+			colors[check.WARN].Printf("== Remediations %s ==\n", r.Type)
 			for _, g := range r.Groups {
 				for _, c := range g.Checks {
 					if c.State == check.FAIL {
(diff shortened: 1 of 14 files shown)


cmd/common.go
```diff
@@ -159,7 +162,7 @@ func prettyPrint(r *check.Controls, summary check.Summary) {
 	// Print remediations.
 	if !noRemediations {
 		if summary.Fail > 0 || summary.Warn > 0 {
-			colors[check.WARN].Printf("== Remediations ==\n")
+			colors[check.WARN].Printf("== Remediations %s==\n", r.Type)
```
lizrice, line 165: Nice addition

cmd/master.go
```diff
@@ -34,7 +34,7 @@ var masterCmd = &cobra.Command{
 		}
 
 		filename := loadConfig(check.MASTER, bv)
-		runChecks(check.MASTER, filename)
+		runChecks(check.MASTER, filename, !isMaster())
```
lizrice, line 37: This should always attempt to run regardless of node type, because the user has explicitly asked for the master tests

cmd/node.go
```diff
@@ -34,7 +34,7 @@ var nodeCmd = &cobra.Command{
 		}
 
 		filename := loadConfig(check.NODE, bv)
-		runChecks(check.NODE, filename)
+		runChecks(check.NODE, filename, false)
```
lizrice, line 37: Same comment as for master, if the user expicitly tries to run the node test cases we should make an attempt

cmd/root.go
```diff
@@ -71,35 +72,35 @@ var RootCmd = &cobra.Command{
...
-			// this a gatekeeper for previous versions
-			valid, err := validTargets(bv, []string{string(check.CONTROLPLANE)}, viper.GetViper())
-			if err != nil {
-				exitWithError(fmt.Errorf("error validating targets: %v", err))
-			}
-			if valid {
-				glog.V(1).Info("== Running control plane checks ==")
-				runChecks(check.CONTROLPLANE, loadConfig(check.CONTROLPLANE, bv))
-			}
+		isMasterNode := isMaster()
```
lizrice, line 75: Can you explain your thinking on changing this logic here? I think I preferred all the decisions about whether to skip a section here in one place, but perhaps I am missing something?

lizrice (commented):
Thanks for your work on this so far! I also commented on #528 with a suggestion about how to present the sections that are skipped.

cmd/root.go
```diff
@@ -86,6 +87,8 @@ var RootCmd = &cobra.Command{
 				glog.V(1).Info("== Running control plane checks ==")
 				runChecks(check.CONTROLPLANE, loadConfig(check.CONTROLPLANE, bv))
 			}
+		} else {
+			fmt.Println("== Skipping master checks ==")
```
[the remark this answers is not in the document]
lizrice: Let's keep these as log output

lizrice (changes requested):
I'd just like to change some of the stdout output into logs (see the suggestions) but otherwise this looks good, thank you!

lizrice:
@yoavrotems @danielpacak note that when we merge this, it will change the default JSON output to include overall totals

czunker:
Hi @lizrice ,
changed the output according to your suggestions.

yoavrotems:
@czunker Hey I fixed some conflicts and now the PR seems to fail the tests

czunker:
Hi @yoavrotems ,

rebased to master and fixed the first issue.
It seems, there was a closing bracket missing:
https://github.com/aquasecurity/kube-bench/pull/759/commits/75e5aa9c0a4b5f260b7a1038f5775d31e703e78d#diff-f3346ecefc99144b92b77b4e6ae336aafbd52a75b9587e026ce0325162fc7afdR709

No some results fail in the integration tests. Correcting the numbers. That just than pass the integration test.

lizrice (approved):
lgtm

PR #760, opened 2020-10-30
Fix typos and formatting

valerianpereira:
Fix typos and formatting


yoavrotems:
Thanks for contributing, but this is no type since we don't talk about checks as a word but on multiple object from 'check' type.
So I'm closing it, you are more then welcome to look through our issues and help out there ! :)

PR #761, opened 2020-11-02
Customize datadir locations for etcd

### LINKED ISSUE #583: The etcd data directory related checkes support relative path
https://github.com/aquasecurity/kube-bench/blob/77f66511e70c4b2c1f83988786e3de04ec7349db/cfg/cis-1.5/master.yaml#L283-L292

How about support the case that the value of `--data-dir` is relative path:

```
etcd --data-dir=data.etcd ...
```

mozillazg:
Use this way to support blow cases:

* no `--data-dir` argument, etcd will use default value(`${name}.etcd`)  as data-dir
* value of `--data-dir` is a relative path. Closes #583

refer:

* https://etcd.io/docs/v3.4.0/op-guide/configuration/#--data-dir


lizrice:
Sorry @mozillazg, this closed inadvertently when we moved master branch to main :-(

PR #762, opened 2020-11-02
Support case which run etcd as systemd service instead of pod

mozillazg:


yoavrotems (approved):
LGTM Thank you @mozillazg

PR #764, opened 2020-11-06
update in README.md

rishabh710:
Addition of table of content

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,3 +1,10 @@
+# Table of content
+-[Test and config files]
+  -[Controls]
+  -[Groups]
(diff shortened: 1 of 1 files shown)


yoavrotems (changes requested):
Hey nice add, we didn't notice it missing, but if we going to add TOC then we not in the same format as the main readme? https://github.com/aquasecurity/kube-bench/blob/master/README.md#table-of-contents

yoavrotems:
Hey I want to close this PR in the next few day :) so would like to get a ping from you if you want to finish it 
@rishabh710

Issue #743, opened 2020-10-14
Suggest revising test 1.3.6 to ensure RotateKubeletServerCertificate is not found or not false


jeffnai-github:
By default, kube-apiserver sets [RotateKubeletServerCertificate=true](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/#options) In test 1.3.6, instead of looking for "RotateKubeletServerCertificate=true", I'd like to suggest the following:

      - id: 1.3.6
        text: "Ensure that the RotateKubeletServerCertificate argument is set to true (Automated)"
        audit: "/bin/ps -ef | grep $controllermanagerbin | grep -v grep"
        tests:
          bin_op: or
          test_items:
            - flag: "--feature-gates"
              compare:
                op: nothave
                value: "RotateKubeletServerCertificate=false"
            - flag: "--feature-gates"
              set: false

mozillazg:
Yes, you are right.

From https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/

<img width="601" alt="image" src="https://user-images.githubusercontent.com/485054/98821429-8287c000-246a-11eb-9bc0-f8432d5b6913.png">

From https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
<img width="752" alt="image" src="https://user-images.githubusercontent.com/485054/98821476-93d0cc80-246a-11eb-86b6-796efe89ec72.png">

From CIS:
<img width="898" alt="image" src="https://user-images.githubusercontent.com/485054/98821597-ba8f0300-246a-11eb-9652-0adb1ed572fc.png">

PR #767, opened 2020-11-11
Pass RotateKubeletServerCertificate related checks if it's not found

mozillazg:
Closes https://github.com/aquasecurity/kube-bench/issues/743

* https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/
* https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/
* https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/


yoavrotems (approved):
LGTM ! 
Just wanted to confirm since when is this change to make sure we don't over look versions where is not set to true by default. 
https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-alpha-or-beta-features 
In here you can see that its true since beta 1.12 sooooo all is good :)

PR #779, opened 2020-12-17
Customize datadir locations for etcd

### LINKED ISSUE #583: The etcd data directory related checkes support relative path

mozillazg:
Use this way to support blow cases:

* no `--data-dir` argument, etcd will use default value(`${name}.etcd`)  as data-dir
* value of `--data-dir` is a relative path. Closes #583

refer:

* https://etcd.io/docs/v3.4.0/op-guide/configuration/#--data-dir
* https://github.com/aquasecurity/kube-bench/pull/761


mozillazg:
ready for review

mozillazg:
@yoavrotems ping ~ 😸

lizrice:
I'd rather see all the variable substitution being done in the Go code rather than as scripts in the test YAML - that way the scripts are simpler, they stay closer to what's described in the benchmark, and we don't need to repeat similar substitution code in several tests. 

We would probably need to extend the config.yaml syntax so that we can indicate that for a given binary ($etcdbin), we want to look for the value of a certain parameter (--data-dir), and substitute that into a given variable name ($etcdatadir)

yoavrotems:
Ping @mozillazg Hey I just want to clean a bit the project, so are you going to comment or commit about liz suggestions? Or should I close this PR and just open an issue in the subject? 
Thank you for the work you have already done.

mozillazg:
@yoavrotems I have no idea how to extend the config.yaml syntax. Could you give an example?

PR #781, opened 2020-12-17
Fix AWS ECR authentication docs

gregdek:
The command you listed here did not work. The command from the official documentation did:

https://docs.aws.amazon.com/AmazonECR/latest/userguide/getting-started-cli.html

`aws ecr get-login-password --region region | docker login --username AWS --password-stdin aws_account_id.dkr.ecr.region.amazonaws.com`

Thanks!


lizrice (approved):
oh yes, you are absolutely right - thanks!

PR #783, opened 2020-12-21
Refactor group skip

lizrice:
Dupe of #776 after moving master to main

PR #785, opened 2020-12-21
Restore changes missed when moving master branch to main 

lizrice:
Restore changes to main that I missed when I used an old version of the master branch 😭


lizrice:
Ended up merging this set of changes with #783 because Travis was being uncooperative

PR #786, opened 2020-12-21
Create pointless main.yml

lizrice:
Trying to force GitHub Actions to appear


lizrice:
Realized we had the wrong name for the workflows directory and that's why Actions weren't showing up

PR #790, opened 2020-12-23
chore: build and push action for ecr and docker

krol3:
- Remove the publish-ecr file 
- Use the build-push action in publish.yml with docker and ecr.
- Change the name secret for the use in the organization (DOCKER_HUB_USERNAME)
- Build using only platforms: linux/amd64, less time to build it.


yoavrotems (approved):
LGTM

lizrice:
I'm not sure about 386 platform, but I think we do need the arm build as well as and

PR #794, opened 2021-01-10
Refactor of EKS and ASFF integration Job and instructions

mkandelaars:
PR with some improvements for deploying kube-bench on EKS and sending findings to Security Hub.

Created an issue here #791, but the summary of the changes in this PR is...
1. Created new Job for EKS/ASFF `job-eks-asff.yaml`, which contains a ConfigMap with the custom configuration for your cluster. This means the configuration is not baked into the image and therefore the image does not need to rebuilt for each Kubernetes cluster / environment.
2. Added a ServiceAccount to the new `job-eks-asff.yaml` Job to make it easier for users when using IAM Roles for Service Accounts. 
3. Removed ASFF instructions from the original `job-eks.yaml` and instead refer to `job-eks-asff.yaml` when sending findings to AWS Security Hub.
4. Reworded the documentation for ASFF with how to use the new Job and removed instructions for rebuilding the image with baked in configuration.


docs/asff.md
```diff
@@ -36,16 +36,17 @@ Here is an example IAM Policy that you can attach to your EKS node group's IAM R
 }
 ```
 
-## Configure and rebuild kube-bench
+## Build kube-bench
 
-You will need to download, build and push the kube-bench container image to your ECR repo as described in Step 3 of the [EKS instructions][eks-instructions], except that before you build the container image, you need to edit `cfg/eks-1.0/config.yaml` to specify the AWS account, AWS region, and the EKS Cluster ARN.
+You will need to download, build and push the kube-bench container image to your ECR repo as described in Step 3 of the [EKS instructions][eks-instructions].
```
lizrice, line 41:
```suggestion
You will need to download, build and push the kube-bench container image to your ECR repo as described in Step 3 of the [EKS instructions][eks-instructions]. Before you build the container image, you need to edit the job configuration.
```

docs/asff.md
```diff
@@ -36,16 +36,17 @@ Here is an example IAM Policy that you can attach to your EKS node group's IAM R
...
 
 ## Modify the job configuration
```
lizrice, line 43:
```suggestion
### Modify the job configuration
```

lizrice (changes requested):
Looks great, thanks for this! 

I think we should remove the three environment variable placeholders from cfg/eks-1.0/config.yaml, would you agree? 

Also made a couple of tiny doc suggestions

docs/asff.md
mkandelaars: @lizrice Not sure on adding that line as the job configuration doesn't need to be added before the container image has been built / pushed. But now with the v0.5.0 release, this whole block is probably redundant now as the image doesn't need to be custom built anymore - so maybe we could remove this section entirely. What do you think?

lizrice: That's true, let's get rid of this part

docs/asff.md
```diff
@@ -36,16 +36,17 @@ Here is an example IAM Policy that you can attach to your EKS node group's IAM R
...
 
-You will need to download, build and push the kube-bench container image to your ECR repo as described in Step 3 of the [EKS instructions][eks-instructions], except that before you build the container image, you need to edit `cfg/eks-1.0/config.yaml` to specify the AWS account, AWS region, and the EKS Cluster ARN.
+You will need to download, build and push the kube-bench container image to your ECR repo as described in Step 3 of the [EKS instructions][eks-instructions].
 
-## Modify the job configuration
+### Modify the job configuration
 
-* Modify `job-eks.yaml` to specify the `--asff` flag, so that kube-bench writes output in ASFF format to Security Hub
-* Make sure that `job-eks.yaml` specifies the container image you just pushed to your ECR registry.
+* Modify the kube-bench Configmap in `job-eks-asff.yaml` to specify the AWS account, AWS region, and the EKS Cluster ARN.
```
lizrice, line 41:
```suggestion
* Modify the kube-bench Configmap in `job-eks-asff.yaml` to specify the AWS account, AWS region, and the EKS Cluster ARN.
* In the same file, modify the image specifed in the Job to use the kube-bench image pushed to your ECR
```

lizrice (approved):

Issue #634, opened 2020-07-05
Search config or bin file via ps when possible


mozillazg:
TODO:

* [ ] `$apiserverbin`
* [ ] `$apiserverconf`
* [ ] `$controllermanagerbin`
* [ ] `$controllermanagerconf`
* [ ] `$etcdbin`
* [ ] `$etcdconf`
* [ ] `$kubeletbin`
* [ ] `$kubeletcafile`
* [ ] `$kubeletconf`
* [ ] `$kubeletkubeconfig`
* [ ] `$kubeletsvc`
* [ ] `$nodesvc`
* [ ] `$proxykubeconfig`
* [ ] `$proykubeconfig`
* [ ] `$schedulerbin`
* [ ] `$schedulerconf`

PR #795, opened 2021-01-11
Try to search the right kubeconfig files and service files

DOboznyi:
To have better support of kops clusters we should change the approach for searching config files.

Ref #634

### DIFF
--- a/cfg/cis-1.6/master.yaml
+++ b/cfg/cis-1.6/master.yaml
@@ -196,7 +196,10 @@ groups:
 
       - id: 1.1.15
         text: "Ensure that the scheduler.conf file permissions are set to 644 or more restrictive (Automated)"
-        audit: "/bin/sh -c 'if test -e $schedulerkubeconfig; then stat -c permissions=%a $schedulerkubeconfig; fi'"
+        audit: |
+          KUBECONFIG=$(ps -ef | grep kube-scheduler | grep -- --config= | awk -F '--config=' '{print $2}' | awk '{print $1}')
+          if test -z $KUBECONFIG; then KUBECONFIG=$schedulerkubeconfig; fi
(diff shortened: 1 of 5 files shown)


DOboznyi:
@lizrice @yoavrotems could you review this PR?

cfg/cis-1.6/master.yaml
```diff
@@ -206,24 +209,30 @@ groups:
         remediation: |
           Run the below command (based on the file location on your system) on the master node.
           For example,
-          chmod 644 $schedulerkubeconfig
+          chmod 644 $KUBECONFIG
```
yoavrotems, line 212: I think this will not since variables sets in audit won't be saved, currently I think it will be best to edit the config file and add there more paths, It will be a good thing the add more dynamically tests for configs, but its need to be part of the code and not in the audit

yoavrotems (changes requested):

yoavrotems:
Thanks for the contribution, pay attention to which branch you are trying to merge, this is the master branch which got deprecated (We deleted and restored it because we had some issues, but it will be deleted) the head branch now is main

cfg/cis-1.6/master.yaml
DOboznyi: But such a way we should add an extra path each time to the config. Is it better than getting the path dynamic?

DOboznyi:
@yoavrotems i'll open new PR to main.

Algoss:
@DOboznyi 
I am running kube-bench pod in kubernetes cluster. It is not able to read the correct config, kubeconfig, ca files. Instead it is reading the default files for each. Hence, pod logs are giving false positive results.

Ref:
https://github.com/aquasecurity/kube-bench/issues/1256

Please provide your input.

PR #797, opened 2021-01-12
Automate 1.2.1

DOboznyi:
Currently, 1.2.1 has automated checks but in results it always warns. So without a manual type, it will fail or pass automatically.


DOboznyi:
@lizrice @yoavrotems please review this PR.

yoavrotems (approved):
LGTM thanks

PR #798, opened 2021-01-13
Fix 1.1.7 1.1.8 for KOPS

DOboznyi:
In KOPS used etc-manager which bootstrap 2 etcd clusters: main (`/etc/kubernetes/manifests/etcd-main.manifest`) and event (`/etc/kubernetes/manifests/etcd-events.manifest`). So i changed logic and it check all manifests which contain etcd in naming in `/etc/kubernetes/manifests`.
Signed-off-by: Dmytro Oboznyi <dmytro.oboznyi@syncier.com>

### DIFF
--- a/cfg/cis-1.6/master.yaml
+++ b/cfg/cis-1.6/master.yaml
@@ -91,7 +91,8 @@ groups:
 
       - id: 1.1.7
         text: "Ensure that the etcd pod specification file permissions are set to 644 or more restrictive (Automated)"
-        audit: "/bin/sh -c 'if test -e $etcdconf; then stat -c permissions=%a $etcdconf; fi'"
+        audit: "/bin/sh -c 'if test -e $etcdconf; then find $etcdconf -name '*etcd*' | xargs stat -c permissions=%a; fi'"
+        use_multiple_values: true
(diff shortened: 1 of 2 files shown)


DOboznyi:
@yoavrotems Could you review this PR?

yoavrotems:
Its a bit too restrictive to my opinion, we are using config.yaml to be less fixed and more flexible, but now it could cause to someone who use config.yaml to add is own custom yaml that doesn't include etcd in the name or even include etcd but has caps "myEtcdForDev.yaml" it will fail the test.
better solution could be to maybe add option to have "AND" in config.yaml currently its like "OR", having AND could work with it like "use_multiple_values" flag, then all the standard config will still and the custom won't fail because we check for a specific pattern in file name.

yoavrotems (changes requested):
Look at the comment above ^

DOboznyi:
The main problem for me that kube-bench test only first found entry from configuration file and i can't there some pattern for search...

yoavrotems (approved):
Ok I see what you saying, since it doesn't effect any existing values in config.yaml it will be fine to merge it

PR #812, opened 2021-01-31
deprecate master and node subcommands

### LINKED ISSUE #810: Deprecate master & node subcommands
We have subcommands `master` and `node` that reflect old versions of the benchmark where there were just two sections, which were implemented in test files `cfg/<benchmark version>/master.yaml` and `cfg/<benchmark version>/node.yaml` respectively. For all the spec versions we now support, there are several more sections (for example etcd has its own section). We have the `--targets` parameter for specifying which test sections to run. 

We should mark the `master` and `node` subcommands as deprecated 
 - add this to the help 
 - make sure the documentation doesn't include examples with these subcommands
 - log a message if either of these subcommands are invoked

A future release should retire them altogether

neha-viswanathan:
Closes #810

### DIFF
--- a/cmd/master.go
+++ b/cmd/master.go
@@ -37,7 +37,7 @@ var masterCmd = &cobra.Command{
 		runChecks(check.MASTER, filename)
 		writeOutput(controlsCollection)
 	},
-	Deprecated: "this command will be retired soon. Please use the `run` command with `--targets="master"` instead.",
+	Deprecated: "this command will be retired soon. Please use the `run` command with `--targets=master` instead.",
 }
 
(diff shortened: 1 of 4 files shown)


yoavrotems (changes requested):
Hey thanks for the contribution! 
I would suggest also to add what is the alternative:
use the run command with --targets="master" or --targets="node" and also need to update the integration testing which is currently failing because it uses those commands (master|node) and doesn't expect the deprecation note in the output

yoavrotems (approved):
LGTM :) thank you for contributing

Issue #811, opened 2021-01-30
[0.5.0] The main config.yaml file is missing


rdbraber:
**What steps did you take and what happened:**

Download the tar.gz file: https://github.com/aquasecurity/kube-bench/releases/download/v0.5.0/kube-bench_0.5.0_linux_arm64.tar.gz. When checking the content:
```
tar -tvf kube-bench_0.5.0_linux_amd64.tar.gz
-rw-r--r-- runner/docker    77 2021-01-20 13:58 cfg/aks-1.0/config.yaml
-rw-r--r-- runner/docker  1323 2021-01-20 13:58 cfg/aks-1.0/controlplane.yaml
-rw-r--r-- runner/docker 13979 2021-01-20 13:58 cfg/aks-1.0/managedservices.yaml
-rw-r--r-- runner/docker    87 2021-01-20 13:58 cfg/aks-1.0/master.yaml
-rw-r--r-- runner/docker 19688 2021-01-20 13:58 cfg/aks-1.0/node.yaml
-rw-r--r-- runner/docker 10809 2021-01-20 13:58 cfg/aks-1.0/policies.yaml
-rw-r--r-- runner/docker    77 2021-01-20 13:58 cfg/cis-1.5/config.yaml
-rw-r--r-- runner/docker  1188 2021-01-20 13:58 cfg/cis-1.5/controlplane.yaml
-rw-r--r-- runner/docker  4653 2021-01-20 13:58 cfg/cis-1.5/etcd.yaml
-rw-r--r-- runner/docker 42541 2021-01-20 13:58 cfg/cis-1.5/master.yaml
-rw-r--r-- runner/docker 20582 2021-01-20 13:58 cfg/cis-1.5/node.yaml
-rw-r--r-- runner/docker  9452 2021-01-20 13:58 cfg/cis-1.5/policies.yaml
-rw-r--r-- runner/docker    77 2021-01-20 13:58 cfg/cis-1.6/config.yaml
-rw-r--r-- runner/docker  1181 2021-01-20 13:58 cfg/cis-1.6/controlplane.yaml
-rw-r--r-- runner/docker  4919 2021-01-20 13:58 cfg/cis-1.6/etcd.yaml
-rw-r--r-- runner/docker 40638 2021-01-20 13:58 cfg/cis-1.6/master.yaml
-rw-r--r-- runner/docker 20048 2021-01-20 13:58 cfg/cis-1.6/node.yaml
[...]

PR #813, opened 2021-01-31
(fix) add config.yaml to releases, see #811

curx:
The main config.yaml is missing at the tar.gz and debian packages.

Signed-off-by: Thorsten Schifferdecker <ts@systs.org>


yoavrotems (approved):
LGTM thank you for contributing :)

PR #815, opened 2021-02-06
chore: fix an invalid markdown syntax

oke-py:
It may be supposed to be a link.


yoavrotems (approved):
Yes thank you for this fix :)

PR #821, opened 2021-02-19
chore: improve bug report template

lizrice:
Ask for more details that we usually need to request anyway when an issue is reported, like the list of currently running processes


yoavrotems (approved):
LGTM

PR #822, opened 2021-02-22
Add logging

yoavrotems:
Some old logging was a bit misleading and lacking in information so I changed some and mostly add more logs, 
Also now the checks logs has the check ID so it much more clear what log belong to which test. 
In the image you can see how a test log looks like now:

image

And this is before:
image

### DIFF
--- a/check/test.go
+++ b/check/test.go
@@ -126,6 +126,9 @@ func (t flagTestItem) findValue(s string) (match bool, value string, err error) 
 		// flag: somevalue
 		// --flag
 		// somevalue
+		// DOESN'T COVER
+		// flag:
+		//	 - wehbook
 		pttn := `(` + t.Flag + `)(=|: *)*([^\s]*) *`
 		flagRe := regexp.MustCompile(pttn)
 		vals := flagRe.FindStringSubmatch(s)
@@ -145,7 +148,7 @@ func (t flagTestItem) findValue(s string) (match bool, value string, err error) 
 			err = fmt.Errorf("invalid flag in testItem definition: %s", s)
 		}
 	}
-	glog.V(3).Infof("In flagTestItem.findValue %s, match %v, s %s, t.Flag %s", value, match, s, t.Flag)
+	glog.V(3).Infof("In flagTestItem.findValue %s", value)
 
 	return match, value, err
 }
@@ -183,6 +186,7 @@ func (t envTestItem) findValue(s string) (match bool, value string, err error) {
 			value = ""
 		}
 	}
+	glog.V(3).Infof("In envTestItem.findValue %s", value)
 	return match, value, nil
 }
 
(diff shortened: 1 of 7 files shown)


check/test.go
```diff
@@ -126,6 +126,9 @@ func (t flagTestItem) findValue(s string) (match bool, value string, err error)
 		// flag: somevalue
 		// --flag
 		// somevalue
+		// DOESN'T COVER
+		// flag:
+		//	 - wehbook
```
lizrice, line 131:
```suggestion
		// DOESN'T COVER - use pathTestItem implementation of findValue() for this
		// flag:
		//	 - wehbook
```

check/test.go
```diff
@@ -232,7 +236,22 @@ func (t testItem) evaluate(s string) *testOutput {
...
+	}
+	if t.auditUsed != "auditCommand"{
+		if t.auditUsed != "auditConfig"{
+			glog.V(3).Info(fmt.Sprintf("Env '%v'%s exist", t.Env, isExist))
+		} else {
+			glog.V(3).Info(fmt.Sprintf("Path '%v'%s exist", t.Path, isExist))
+		}
+	} else {
+		glog.V(3).Info(fmt.Sprintf("Flag '%v'%s exist", t.Flag, isExist))
+	}
```
lizrice, line 251: I think this would be nicer as a switch something like this: 

```suggestion
    switch t.auditUsed {
        case "auditCommand": 
		glog.V(3).Info(fmt.Sprintf("Flag '%v' %s exist", t.Flag, isExist))
        case "auditConfig":
		glog.V(3).Info(fmt.Sprintf("Path '%v' %s exist", t.Flag, isExist))
        ...
    }
```

As an aside, generally speaking if you're using if tests try to test on the positive sense so it's easier to follow the logic: `if t.auditUsed == "auditCommand"` is easier to hold in your head than `if t.auditUsed != "auditCommand"`. But I think a switch will be neater here anyway

check/test.go
```diff
@@ -232,7 +236,22 @@ func (t testItem) evaluate(s string) *testOutput {
 	}
 
 	result.flagFound = match
-	glog.V(3).Info(fmt.Sprintf("found %v", result.flagFound))
+	var isExist string
+	if !result.flagFound{
+		isExist = " not"
+	}
```
lizrice, line 242: Personally I would just trace out "exists: true" or "exists: false" so that you can do it by formatting the boolean value of `result.flagFound`. 

But if you want to make it more of a sentence, for grammatical accuracy it's "X exists" or "X does not exist".  So you could do something like 

```suggestion
	var isExist = "exists"
	if !result.flagFound{
		isExist = "does not exist"
	}
```

and then drop the word "exist" from the format strings later

lizrice (approved):
I made a couple of suggestions but this is fine once you have addressed those

itaysk (approved):

PR #825, opened 2021-02-22
chore: fix defer func in for-loop

dulltz:
`defer` executes at function return, not out of block scope, so we should wrap `for` block by function.


lizrice (approved):
good catch! LGTM

PR #827, opened 2021-02-22
docs: remove deprecated subcommands master / node

lizrice:
Example commands should use auto-detection or the new `run --targets` approach to specifying sections of the benchmark to run


yoavrotems (approved):
LGTM

PR #865, opened 2021-05-03
cfg: add `/etc/default/kubeletconfig.json` for AKS

bcho:
AKS stores [kubelet configuration](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/) JSON in path `/etc/default/kubeletconfig.json` ([ref](https://github.com/Azure/AgentBaker/blob/11d2ec484d6142e996291d6a5a4d61820214eac8/parts/linux/cloud-init/artifacts/cse_config.sh#L207)).

To make kube-bench works for AKS clusters using kubelet config, this pull request added this file into kubelet configuration files list.


cfg/config.yaml
```diff
@@ -147,6 +147,7 @@ node:
       - "/etc/kubernetes/kubelet/kubelet-config.json"
       - "/home/kubernetes/kubelet-config.yaml"
       - "/home/kubernetes/kubelet-config.yml"
+      - "/etc/default/kubeletconfig.json"
```
bcho, line 150: FYI: the order of these files matters, so here we expect to check with `/etc/default/kubeletconfig.json` first, then `/etc/default/kubetlet`, since both files present in AKS cluster. But I am not 100% sure this is the correct way to do this.

yoavrotems: Since the is no way to priorities beside tops to bottom this is the correct way, I actually didn't find any information online beside the link you posted in github, but there is no harm in adding new paths :) after the build will run I will approve it

Issue #867, opened 2021-05-04
Seeing "No such file or directory" for CIS benchmark 1.1.9 using K8s 1.2.1


davidhay1969:
**Overview**

I'm seeing: -

```text
 "permissions=644\npermissions=644\nfind: /var/lib/cni/networks: No such file or directory\n"
```

as part of the output from `kube-bench` for CIS test `1.1.9` ( and also for `1.1.10` ).

This is being run against the K8s Master ( Control Plane ) node.

More specifically, this is the entire output from that particular run: -

```text
I0504 17:36:02.657095   21516 common.go:343] Mapped Kubernetes version: 1.21 to Benchmark version: cis-1.6
I0504 17:36:02.657102   21516 common.go:346] Kubernetes version: "1.21" to Benchmark version: "cis-1.6"
...
I0504 17:36:02.705809   21516 check.go:110] -----   Running check 1.1.9   -----
I0504 17:36:02.712997   21516 check.go:299] Command: "ps -ef | grep kubelet | grep -- --cni-conf-dir | sed 's%.*cni-conf-dir[= ]\\([^ ]*\\).*%\\1%' | xargs -I{} find {} -mindepth 1 | xargs stat -c permissions=%a\nfind /var/lib/cni/networks -type f | xargs --no-run-if-empty stat -c permissions=%a"
I0504 17:36:02.713018   21516 check.go:300] Output:
 "permissions=644\npermissions=644\nfind: /var/lib/cni/networks: No such file or directory\n"
I0504 17:36:02.713091   21516 check.go:221] Running 1 test_items
I0504 17:36:02.713152   21516 test.go:151] In flagTestItem.findValue 644
I0504 17:36:02.713161   21516 test.go:245] Flag 'permissions' exists
I0504 17:36:02.713217   21516 test.go:151] In flagTestItem.findValue 644
I0504 17:36:02.713226   21516 test.go:245] Flag 'permissions' exists
I0504 17:36:02.713259   21516 test.go:151] In flagTestItem.findValue 
I0504 17:36:02.713265   21516 test.go:245] Flag 'permissions' does not exist
I0504 17:36:02.713301   21516 check.go:245] Used auditCommand
[...]

PR #877, opened 2021-05-20
Issue 867: Fixing "No such file or directory" issue seen with CIS 1.1.9 and 1.1.0

davidhay1969:
This PR covers the issue documented in #867 whereby CIS 1.1.9 ( and also 1.1.10 ) throw a WARNing for `/var/lib/cni/networks` in terms of permissions ( `chown` and `chmod` ), even though the same command, when run locally on the K8s nodes does not return an error.

#867 describes the issue in more detail but, in short, we see: -

```text
I0504 17:36:02.705809   21516 check.go:110] -----   Running check 1.1.9   -----
I0504 17:36:02.712997   21516 check.go:299] Command: "ps -ef | grep kubelet | grep -- --cni-conf-dir | sed 's%.*cni-conf-dir[= ]\\([^ ]*\\).*%\\1%' | xargs -I{} find {} -mindepth 1 | xargs stat -c permissions=%a\nfind /var/lib/cni/networks -type f | xargs --no-run-if-empty stat -c permissions=%a"
I0504 17:36:02.713018   21516 check.go:300] Output:
 "permissions=644\npermissions=644\nfind: /var/lib/cni/networks: No such file or directory\n"
```

in the verbose `kube-bench` logs which results in WARNings for both `1.1.9` and `1.1.10` even though permissions and ownership of `/var/lib/cni/networks` are A-OK.

The issue relates to the second audit command - `    find /var/lib/cni/networks -type f | xargs --no-run-if-empty stat -c permissions=%a` - which returns "No such file or directory", but *only* when run via `kube-bench`.

The mitigation is to add output redirection - `2> /dev/null` to the `find` command.

This has been most recently tested using K8s `1.21` on Linux on IBM Z ( `s390x` ), by manually amending `cfg/cis-1.6/master.yaml` and building a new `kube-bench` image.

Signed-off by: Dave Hay <david_hay@uk.ibm.com>

### DIFF
--- a/cfg/cis-1.6/master.yaml
+++ b/cfg/cis-1.6/master.yaml
@@ -122,7 +122,7 @@ groups:
         text: "Ensure that the Container Network Interface file permissions are set to 644 or more restrictive (Manual)"
         audit: |
           ps -ef | grep $kubeletbin | grep -- --cni-conf-dir | sed 's%.*cni-conf-dir[= ]\([^ ]*\).*%\1%' | xargs -I{} find {} -mindepth 1 | xargs stat -c permissions=%a
-          find /var/lib/cni/networks -type f | xargs --no-run-if-empty stat -c permissions=%a
+          find /var/lib/cni/networks -type f 2> /dev/null | xargs --no-run-if-empty stat -c permissions=%a
         use_multiple_values: true
         tests:
           test_items:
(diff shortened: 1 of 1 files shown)


yoavrotems:
Thanks for contributing and helping out! 
LGTM

PR #886, opened 2021-06-04
False positive when running rh-0.7 benchmarks

tonyqui:
Got few false positive alerts on a kube-bench report due to some typos.


yoavrotems (approved):
LGTM, thank you for contributing

Issue #891, opened 2021-06-08
Image not supporting jq


yoavrotems:
In old kube-bench cis version there wasn't any use in jq, now in openshift 4.X hardening guide there is a lot of use in jq and it not installed or mounted to image.

Issue #890, opened 2021-06-08
Job can't run properly on openshift


yoavrotems:
**Overview**
 
Openshift 4 + CIS is using the oc command 
oc can't run on simple Alpine since it use a glibc and its not in Alpine. 

**How did you run kube-bench?**

oc create -f job.yaml
**What happened?**

Getting 
 I0526 12:50:23.675148  989968 util.go:498] Can't use oc command: fork/exec /usr/local/mount-from-host/bin/oc: no such file or directory

**What did you expect to happen:**

I expect kube-bench to auto detect ocp4.6 and run it. 
**Environment** 

openshift:
Client Version: 4.6.0-202104300142.p0.git.5ab7b2b-5ab7b2b
Server Version: 4.6.30
Kubernetes Version: v1.19.0+9caf8fe
Client Version: 4.6.0-202104300142.p0.git.5ab7b2b-5ab7b2b
Server Version: 4.6.30
Kubernetes Version: v1.19.0+9caf8fe
kube-bench:0.6.2

**Configuration files**

rh-0.1

**Anything else you would like to add:**

[Miscellaneous information that will assist in solving the issue.]

PR #892, opened 2021-06-08
Fix ocp job issues

yoavrotems:
There is two issues with running job on openshift, first issue [Alpine not supporting glibc](https://github.com/aquasecurity/kube-bench/issues/890) and the the second issue [jq is not mounted or installed for ocp4.X](https://github.com/aquasecurity/kube-bench/issues/891)


yoavrotems:
Accidently included unrelate commits, I will close it and open a new one with only the needed commits.

PR #893, opened 2021-06-08
Fix ocp job issues

yoavrotems:
There is two issues with running job on openshift, first issue [Alpine not supporting glibc](https://github.com/aquasecurity/kube-bench/issues/890) and the the second issue [jq is not mounted or installed for ocp4.X](https://github.com/aquasecurity/kube-bench/issues/891)

Also Update build.yaml to only build goreleaser and not to release when push to main.


Dockerfile
```diff
@@ -20,6 +20,12 @@ RUN apk --no-cache add procps
 # https://github.com/aquasecurity/kube-bench/issues/535
 RUN apk --no-cache add openssl
 
+# Add glibc for running oc command 
+RUN wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://alpine-pkgs.sgerrand.com/sgerrand.rsa.pub
+RUN wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/2.33-r0/glibc-2.33-r0.apk
+RUN apk add glibc-2.33-r0.apk
+RUN apk add jq
```
danielpacak, line 27: Is jq required by kube-bench or oc binary?

danielpacak (approved):
LGTM!

Dockerfile
yoavrotems: by kube-bench, but it currently only in use in cis openshift 4

Issue #903, opened 2021-06-11
"Message:Finding does not adhere to Amazon Finding Format"


stephanlindauer:
With version `0.6.2` i get recommendation text that doesn't fit the requirements of that field in the ASFF spec:
```
Number of findings that failed to import:2
Message:Finding does not adhere to Amazon Finding Format. data.Remediation.Recommendation.Text should NOT be longer than 512 characters.
Error Code:InvalidInput
```

yoavrotems:
Could you add more information? 
Here you can see what information we need https://github.com/aquasecurity/kube-bench/blob/main/CONTRIBUTING.md#bugs

yoavrotems:
Also would help to know what is the latest version you used that it worked since eks files wasn't changed for 7 months

yoavrotems:
@stephanlindauer Do you might ran a different cfg?

stephanlindauer:
thanks!

adelwin:
I'm not sure if there's a regression or if something in AWS-SH has added some further checking.
I am getting something similar to this one.
```
2022/01/26 06:31:11 ID:arn:aws:securityhub:ap-southeast-1::product/aqua-security/kube-bench421940619307EKSnodeID+4.1.42022-01-26T06:31:11Z
2022/01/26 06:31:11 Message:Finding does not adhere to Amazon Finding Format. data.Remediation.Recommendation.Text should NOT be shorter than 1 characters.
2022/01/26 06:31:11 Error Code:InvalidInput
2022/01/26 06:31:11 Number of findings that failed to import:13
2022/01/26 06:31:11 ID:arn:aws:securityhub:ap-southeast-1::product/aqua-security/kube-bench421940619307EKSnodeID+5.1.12022-01-26T06:31:11Z
2022/01/26 06:31:11 Message:Finding does not adhere to Amazon Finding Format. data.Remediation.Recommendation.Text should NOT be shorter than 1 characters.
```
Should I open a new ticket or continue the convo here?

yoavrotems:
@adelwin yes please open a new bug report with all of the needed information

adelwin:
I just checked that the new image v0.6.6 has already addressed the recommendation string.
Thanks @yoavrotems

yoavrotems:
Sure thing 👍

PR #915, opened 2021-06-20
Check string size

yoavrotems:
ASFF string size need to be constraint, could explain https://github.com/aquasecurity/kube-bench/issues/903
`Message:Finding does not adhere to Amazon Finding Format. data.Remediation.Recommendation.Text should NOT be longer than 512 characters.
Error Code:InvalidInput`


debu99:
missing eks-1.0.1 in the aquasec/kube-bench:latest image

yoavrotems:
It's not missing but just not there yet, it will be released in the next release, docker image used to be created from each push to main, but we changed it to be triggered by new tags. 
Because it used to create unstable images for each new push. If you have any more questions in the subject please open a issue about it. 
Thanks

PR #922, opened 2021-06-24
cis-1.20 section 1.1.10 command revision.

Hacks4Snacks:
An erroneous xargs statement resulting in an empty response when running the audit command.


yoavrotems:
LGTM thanks 👍

PR #923, opened 2021-06-29
K8s Job Command Clean

njgibbon:
Hello,

* Updates the `job-*.yamls` for consistency purposes and updated to use  `run --targets` as the prior statements in-use have been deprecated.

ItAintMuch

Thank you for Kube-bench.

### DIFF
--- a/job-iks.yaml
+++ b/job-iks.yaml
@@ -10,7 +10,7 @@ spec:
       containers:
         - name: kube-bench
           image: aquasec/kube-bench:latest
-          command: ["kube-bench", "run", "--targets", "node", "--version", "1.13"]
+          command: ["kube-bench", "run", "--targets", "node", "--version", "1.20"]
           volumeMounts:
             - name: var-lib-kubelet
               mountPath: /var/lib/kubelet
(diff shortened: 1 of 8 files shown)


job-iks.yaml
```diff
@@ -10,7 +10,7 @@ spec:
       containers:
         - name: kube-bench
           image: aquasec/kube-bench:latest
-          command: ["kube-bench", "--version", "1.13", "node"]
+          command: ["kube-bench", "run", "--targets", "node", "--version", "1.13"]
```
yoavrotems, line 13: We should also update the version here to 1.20 🤙

yoavrotems (changes requested):

yoavrotems:
Beside that command everything looks good and a nice honest work !  😉

job-iks.yaml
njgibbon: Hey - yes, makes sense thank you :) Just updated.

yoavrotems (approved):

yoavrotems:
LGTM great! :)

PR #930, opened 2021-07-07
Adding targets flag

steve-heslouin:
Adding correct flags in order to avoid the depreciation warning


steve-heslouin:
closing as duplicata with existing PR :)

PR #941, opened 2021-07-26
Update Dependencies

mmorel-35:
Supersedes #933
This pr updates gomod and github-actions dependencies
Cachés go and python dependencies. 
I have used dependabot to do that, but can cancel remove it's config if it's undesired


mmorel-35:
I'm surprised that it is failing ! See https://github.com/mmorel-35/kube-bench/actions/runs/1073762386

.github/workflows/build.yml
```diff
@@ -24,12 +24,20 @@ jobs:
           go-version: 1.16
       - name: Checkout code
         uses: actions/checkout@v2
+      - uses: actions/cache@v2
```
yoavrotems, line 27: I think cache is unnecessary in this project

.github/workflows/build.yml
```diff
@@ -24,12 +24,20 @@ jobs:
...
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
       - name: yaml-lint
         uses: ibiqlik/action-yamllint@v3
       - name: Run unit tests
         run: make tests
       - name: Upload code coverage
-        uses: codecov/codecov-action@v1
+        uses: codecov/codecov-action@v2.0.2
```
yoavrotems, line 40: better use major release so we won't have to update each minor release

.github/workflows/mkdocs-deploy.yaml
```diff
@@ -19,6 +20,12 @@ jobs:
         with:
           fetch-depth: 0
           persist-credentials: true
+      - uses: actions/cache@v2
```
yoavrotems, line 23: ditto, seems unnecessary to me in this project

.github/workflows/publish.yml
```diff
@@ -41,7 +41,7 @@ jobs:
           password: ${{ secrets.ECR_SECRET_ACCESS_KEY }}
       - name: Get version
         id: get_version
-        uses: crazy-max/ghaction-docker-meta@v1
+        uses: crazy-max/ghaction-docker-meta@v3.4.1
```
yoavrotems, line 44: ditto about major releases

yoavrotems (changes requested):

.github/workflows/publish.yml
mmorel-35:
```suggestion
        uses: crazy-max/ghaction-docker-meta@v3
```

.github/workflows/build.yml
mmorel-35:
```suggestion
        uses: codecov/codecov-action@v2
```

go.mod
```diff
@@ -3,29 +3,25 @@ module github.com/aquasecurity/kube-bench
...
-	github.com/spf13/viper v1.4.0
-	github.com/stretchr/testify v1.4.0
-	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d // indirect
-	google.golang.org/appengine v1.5.0 // indirect
-	gopkg.in/yaml.v2 v2.2.8
+	github.com/spf13/cobra v0.0.7
+	github.com/spf13/viper v1.8.1
+	github.com/stretchr/testify v1.7.0
+	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/api v0.0.0-20190409021203-6e4e0e4f393b
```
mmorel-35, line 25: There is now a v0.21.3 for k8s.io/api , k8s.io/apimachinery and k8s.io/client-go but it requires a lot more adaptations

yoavrotems: I think then that we should wait for a major release, right now its working and I don't see the value for updating it if it will requires a lot more adaptations. 
Lets cross this bridge when they will release version 1

yoavrotems:
Hey thanks for taking care of my notes, I just updated a bit our pipeline (because kind old version was breaking stuff there haha kind of make the point for your PR) 
I would like you to update the PR and see there isn't any conflicts after this I don't see a reason not to approve it 😃

yoavrotems (approved):
Great! LGTM thank you for contributing

PR #942, opened 2021-07-27
Adding AWS SNS feature

steve-heslouin:
Hello,

In our organisation, in order to efficiently centralise security related content, we heavily rely on SNS -> SQS to achieve cross account/cross regions communications.

So we added this SNS functionality to kube-bench, if you could please take a look and let us know your thoughts whether you would consider adding this functionality into kube-bench, it would (at least for us) of a great help ;)

Thanks


yoavrotems:
Hey looks good, but I don't use SNS so I'm not sure how it suppose to look could you please attach more information about the output? 
Because I see that you just sending the entire json as massage and I know from asff there are size limitations. 
It will be helpful to see how its integrating

steve-heslouin:
@yoavrotems 

Sure, see SNS as a message queue software, in the same idea as RabbitMQ, but made in AWS. It easily integrate with various different AWS service. You can Publish messages to its "topic" from AWS Service or SDK and Subscribe to it to receive the message:

Here is a diagram on a possible use case for Kube-Bench:

Screenshot 2021-08-02 at 12 54 25

Basically, imagine multiple EKS cluster in different accounts/regions that you want to centralise the scan result in DB for compliance reason. You send result from EKS to SNS, on the other side SQS subscribe to it and deliver it to lambda. Your lambda code will parse the json result of the scan and store what needs to be stored in database.

Example of what payload would lambda receive in such case:
```
{
  "Records": [
    {
      "messageId": "uuid",
      "receiptHandle": "receiptHandle"
      "body": "{}"
        "attributes": {
        "ApproximateReceiveCount": "2",
        "SentTimestamp": "SentTimestamp",
        "SenderId": "SenderId",
        "ApproximateFirstReceiveTimestamp": "ApproximateFirstReceiveTimestamp"
      },
      "messageAttributes": {},
      "md5OfBody": "md5OfBody",
      "eventSource": "aws:sqs",
      "eventSourceARN": "arn....",
      "awsRegion": "us-east-1"
    }
  ]
}
```

Where the actual full scan of the results would be in `$.Records[0].body`

I join a concrete example of a test ran on a dummy test EKS cluster.
https://pastebin.com/7uxNUKM5


For the size, by default it can scale up to 256kb, so quite a lot in information that we wouldn't not exceed with any of those tests results imo

yoavrotems:
Our integration test pipeline failed because we used an old version of kind which seems to start breaking 😢 so I changed some stuff (removed integration from go code to run by github action) and now it fixed 🥳 

Thank you for the detailed explanation! it looks good to me and I see the value of it, just missing unit testing, after adding those I will approve this. 
Thank you for taking part and contributing ! 💣

mozillazg:
@steve-heslouin Hi Steve, are you still working on this?

itaysk:
@mozillazg @steve-heslouin while I love the initiative and appreciate the effort, I think this is out of scope for kube-bench. 
These kind of integrations should be done with another tool that consumes the output of kube-bench and process it however is needed. potentially it may be in the same project that aggregates and manages the results that you later described. If you are still interested in this, I would encourage you to take this code into a new project and if you would be interested we would love to mention this tool in the documentation and share an example of how to use kube-bench with SNS using this tool.

PR #951, opened 2021-08-04
Use latest of k8s.io/client-go

mmorel-35:
k8s.io/client-go is in  v11.0.0+incompatible, v0.21.3 is the latest


yoavrotems (changes requested):
Could you ping or assign me because until I approved the first PR, now there are conflict with this PR and the other #952

PR #952, opened 2021-08-05
Use latest of github.com/spf13/cobra

mmorel-35:
github.com/spf13/cobra is in v0.0.7, v1.2.1 is the latest


yoavrotems (changes requested):
Could you ping or assign me because until I approved the first PR, now there are conflict with this PR and the other #951

PR #971, opened 2021-08-18
Add support for CIS benchmarking of Platform9 Managed Kubernetes clusters in kube-bench

pallabpain:
[**Platform9 Managed Kubernetes** ](https://platform9.com/managed-kubernetes/features/) (or PMK) is a production-grade K8s platform since 2017. This PR includes custom configuration files for PMK clusters that are based on `cis-1.20`.

Here's a list of changes that are a part of this PR
1. `config.yaml` has been updated with paths for bins and configs
2. `job-pmk.yaml` has been added for running `pmk` benchmarks
3. Updated `cfg/pmk-1.0/master.yaml` sections `1.1.11`, `1.1.12`, `1.1.13`, `1.1.14`, `1.1.19`, `1.1.20`, `1.1.21` 
4. Updated `cfg/pmk-1.0/node.yaml` sections `4.1.10`

### DIFF
--- a/cfg/config.yaml
+++ b/cfg/config.yaml
@@ -18,6 +18,8 @@ master:
     - kubelet
 
   kubernetes:
+    confs:
+      - "/etc/pf9/kube.d/kubeconfigs/admin.yaml"
     defaultconf: /etc/kubernetes/config
 
   apiserver:
@@ -29,6 +31,7 @@ master:
       - "openshift start master api"
       - "hypershift openshift-kube-apiserver"
     confs:
+      - /etc/pf9/kube.d/master.yaml
       - /etc/kubernetes/manifests/kube-apiserver.yaml
       - /etc/kubernetes/manifests/kube-apiserver.yml
       - /etc/kubernetes/manifests/kube-apiserver.manifest
@@ -45,6 +48,7 @@ master:
       - "scheduler"
       - "openshift start master controllers"
     confs:
+      - /etc/pf9/kube.d/master.yaml
       - /etc/kubernetes/manifests/kube-scheduler.yaml
       - /etc/kubernetes/manifests/kube-scheduler.yml
       - /etc/kubernetes/manifests/kube-scheduler.manifest
@@ -53,6 +57,7 @@ master:
       - /etc/origin/master/scheduler.json
     defaultconf: /etc/kubernetes/manifests/kube-scheduler.yaml
     kubeconfig:
+      - /etc/pf9/kube.d/kubeconfigs/kube-scheduler.yaml
       - /etc/kubernetes/scheduler.conf
       - /var/lib/kube-scheduler/kubeconfig
       - /var/lib/kube-scheduler/config.yaml
@@ -68,6 +73,7 @@ master:
       - "openshift start master controllers"
       - "hypershift openshift-controller-manager"
     confs:
+      - /etc/pf9/kube.d/master.yaml
       - /etc/kubernetes/manifests/kube-controller-manager.yaml
       - /etc/kubernetes/manifests/kube-controller-manager.yml
       - /etc/kubernetes/manifests/kube-controller-manager.manifest
@@ -75,6 +81,7 @@ master:
       - /var/snap/microk8s/current/args/kube-controller-manager
     defaultconf: /etc/kubernetes/manifests/kube-controller-manager.yaml
     kubeconfig:
+      - /etc/pf9/kube.d/kubeconfigs/kube-controller-manager.yaml
       - /etc/kubernetes/controller-manager.conf
       - /var/lib/kube-controller-manager/kubeconfig
     defaultkubeconfig: /etc/kubernetes/controller-manager.conf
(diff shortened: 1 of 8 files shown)


yoavrotems (changes requested):
Hey @pallabpain thanks for your contribution. 
I just have a few questions,
On what documentation did you base this spec? from a quick look to me it seems like CIS, if so what is the extra value to have it in a new spec? why not use the existing cis-1.20? a following to this if there is a value have it as a new spec you will need to add auto detection for PMK. 
To my understanding PMK is just another "layer" to add and not a different kubernetes platform, I don't think there is a need for it own specs, just adding path to config.yaml and a new job-pmk.yaml looks sufficient enough to me.

I don't use PMK so if you have other thoughts I would like to hear

pallabpain:
> Hey @pallabpain thanks for your contribution.
> I just have a few questions,
> On what documentation did you base this spec? from a quick look to me it seems like CIS, if so what is the extra value to have it in a new spec? why not use the existing cis-1.20? a following to this if there is a value have it as a new spec you will need to add auto detection for PMK.
> To my understanding PMK is just another "layer" to add and not a different kubernetes platform, I don't think there is a need for it own specs, just adding path to config.yaml and a new job-pmk.yaml looks sufficient enough to me.
> 
> I don't use PMK so if you have other thoughts I would like to hear

@yoavrotems Thanks for the comment and for reviewing the pull request.

You are correct. This is based on CIS-1.20. However, there are a few differences when it comes to how a standard kubernetes cluster is deployed and how the Platform9 Managed Cluster is deployed. For example, PMK doesn't deploy etcd as a pod or as a system binary. It deploys it as a container and hence, the existing CIS-1.20 test couldn't locate its parameters. Similarly, the permissions on some directories vary from the CIS-1.20 benchmark. Hence, I had to create a separate directory and update the respective yamls to accommodate the changes in addition to the `config.yaml` file.

yoavrotems:
Hey, I looked and compared it to cis 1.20 and except etcd didn't noticed any other changes and in etcd I noticed you just removed the -flag in one test (which I don't get why? it a flag/env/path if one fails it will try the other). 
I think only changes to config.yaml is enough to have the different paths, also about that I'm not sure, because if we will except this PR we would need to maintain it as well, we will need to see what to do with it. 
Anyway here I think you should remove the cfg/pkm-1.0 and from there we can discuss about how to continue

pallabpain:
> Hey, I looked and compared it to cis 1.20 and except etcd didn't noticed any other changes and in etcd I noticed you just removed the -flag in one test (which I don't get why? it a flag/env/path if one fails it will try the other).
> I think only changes to config.yaml is enough to have the different paths, also about that I'm not sure, because if we will except this PR we would need to maintain it as well, we will need to see what to do with it.
> Anyway here I think you should remove the cfg/pkm-1.0 and from there we can discuss about how to continue

@yoavrotems I have updated the PR description with the list of changes. In addition to etcd.yaml, I had to update `master.yaml` as well as `node.yaml`

- Updated cfg/pmk-1.0/master.yaml sections 1.1.11, 1.1.12, 1.1.13, 1.1.14, 1.1.19, 1.1.20, 1.1.21
- Updated cfg/pmk-1.0/node.yaml sections 4.1.10

Can you review the above sections?

I'll take a look at the etcd file change as well to see if I can revert it and make it work with our k8s cluster.

pallabpain:
> > Hey, I looked and compared it to cis 1.20 and except etcd didn't noticed any other changes and in etcd I noticed you just removed the -flag in one test (which I don't get why? it a flag/env/path if one fails it will try the other).
> > I think only changes to config.yaml is enough to have the different paths, also about that I'm not sure, because if we will except this PR we would need to maintain it as well, we will need to see what to do with it.
> > Anyway here I think you should remove the cfg/pkm-1.0 and from there we can discuss about how to continue
> 
> @yoavrotems I have updated the PR description with the list of changes. In addition to etcd.yaml, I had to update `master.yaml` as well as `node.yaml`
> 
> * Updated cfg/pmk-1.0/master.yaml sections 1.1.11, 1.1.12, 1.1.13, 1.1.14, 1.1.19, 1.1.20, 1.1.21
> * Updated cfg/pmk-1.0/node.yaml sections 4.1.10
> 
> Can you review the above sections?
> 
> I'll take a look at the etcd file change as well to see if I can revert it and make it work with our k8s cluster.

@yoavrotems ^ Awaiting your review. Let me know if this looks good.

yoavrotems:
I think that there isn't enough value here now, and supporting another platform is significant for us, we don't want it to break in the next pmk release.. 
This is why we work mostly with CIS because its an official documentation that we can relay on. 
I'm closing this PR. 
BUT if you want from what I see most of the changes you did could be apply with parameter substitution we use in config.yaml and minor changes to cis-1.20 ( for example 1.11 just add path to config.yaml in etcdbin, in 1.14 could move the path `/etc/kubernetes/admin.conf` to be inside config.yaml). 
Those changes which will be more general I could except

Issue #802, opened 2021-01-17
Utility fetches incorrect compliance status for few of the GKE CIS recommendations


ankitsrao:
**What steps did you take and what happened:**

I'm currently working on a POC for which I'm trying to run the kube-bench agent on a GKE cluster.

Command used: kube-bench --benchmark gke-1.0

**What did you expect to happen:**

- Expected output: Correct compliance recommendations for CIS GKE policies

- Actual output: For the below policies, even though the configuration on the GKE cluster is correct, the compliance status returned by kube-bench is failed.

Policy list:

4.2.1 Ensure that the --anonymous-auth argument is set to false
4.2.2 Ensure that the --authorization-mode argument is not set to AlwaysAllow for Kubelet
4.2.3 Ensure that the --client-ca-file argument is set as appropriate for Kubelet
4.2.4 Ensure that the --read-only-port argument is set to 0
4.2.6 Ensure that the --protect-kernel-defaults argument is set to true
4.2.10 Ensure that the --tls-cert-file and --tls-private-key-file arguments are set as appropriate for Kubelet
4.2.12 Ensure that the RotateKubeletServerCertificate argument is set to true

Refer below screenshot of the kube-config file of the cluster:
![image](https://user-images.githubusercontent.com/41519748/104846214-689daa00-58ff-11eb-8289-cf47a68e4857.png)

**Environment** 

GKE Version: 1.16.15-gke.6000

Thanking you in anticipation!

lizrice:
Please could you let us know the output from running `ps -eaf | grep kubelet` on this node? Thanks

nealharris:
FWIW, https://cloud.google.com/kubernetes-engine/docs/concepts/cis-benchmarks says that 4.2.4 and 4.2.6 are expected to fail with an out-of-the-box GKE cluster. It also says 4.2.10 and 4.2.12 have 'equivalent controls', so maybe kube-bench will show FAIL in those cases?

4.2.1, 4.2.2, and 4.2.3 are supposed to pass with a default GKE cluster, though.

sp185503:
@lizrice I'm also seeing these fail.  Here's the requested output from my environment:

```

/home/kubernetes/bin/kubelet --v=2 --cloud-provider=gce \
--experimental-check-node-capabilities-before-mount=true \
--experimental-mounter-path=/home/kubernetes/containerized_mounter/mounter \
[...]

PR #976, opened 2021-08-23
GKE: Fix 4.1.9, skip irremediable checks, add /home/kubernetes mount

nkeenan38:
### Description
GKE benchmarks were returning several false-positives and failing tests that GKE has equivalent controls for.
Issue: #802 

### Changes
1. GKE nodes' kubelet config file may be found at `/home/kubernetes/kubelet-config.yaml`. Modified the GKE kubernetes job to mount `/home/kubernetes` so that this file can be found. Also updated the volume mounts to be read only.
2. The CIS 4.1.9 check would fail for appropriate permissions if they were not exactly `644`. Changed the operation from `eq` to `bitmask` to resolve.
3. Skipped the failing checks that Google provides equivalent controls for (4.2.4, 4.2.6, 4.2.9, 4.2.10). See https://cloud.google.com/kubernetes-engine/docs/concepts/cis-benchmarks#default-values for the justifications for these checks.

### DIFF
--- a/cfg/gke-1.0/node.yaml
+++ b/cfg/gke-1.0/node.yaml
@@ -78,7 +78,7 @@ groups:
             - flag: "permissions"
               set: true
               compare:
-                op: eq
+                op: bitmask
                 value: "644"
         remediation: |
           Run the following command (using the config file location identified in the Audit step)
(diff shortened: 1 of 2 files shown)


yoavrotems:
Hey @nkeenan38 thank you for contributing, lgtm!

PR #983, opened 2021-09-09
Support check results to be output to http-server

algebra2k:
**What type of PR is this?**
Support check results to be output to http-server.

/kind feature

**What this PR does**:
The PR can provide the user with an option to output the results to the http-server, which will send the results to the http-server after the check of the kube-bench is executed. 

Users can use the following command to enable:

```shell
kube-bench  --http-server=true --http-server-url="localhost/kube-bench-result-collector/" --http-server-content-type="application/json"
```


jrkeen:
It's pleasant things.

yoavrotems:
Hey ! thanks for contributing, we will release the new kube-bench version and then I will review this PR

yoavrotems:
I feel like this a bit too extra, we already output json and junit.
Sending those with another script is pretty easy and there is a lot of tools for it (Also can config syslog and many other solutions). If you chose to keep it in your fork I would recommend to remove 2 flags because 3 flags are unnecessary, you should only use one and its the url flag. (because if you use it it means the first flag you had will be true and content type is only json)

PR #1009, opened 2021-09-30
Add various paths as used by Talos

lennardk:
Implements #1008

### DIFF
--- a/cfg/config.yaml
+++ b/cfg/config.yaml
@@ -35,6 +35,7 @@ master:
       - /var/snap/kube-apiserver/current/args
       - /var/snap/microk8s/current/args/kube-apiserver
       - /etc/origin/master/master-config.yaml
+      - /etc/kubernetes/manifests/talos-kube-apiserver.yaml
     defaultconf: /etc/kubernetes/manifests/kube-apiserver.yaml
 
   scheduler:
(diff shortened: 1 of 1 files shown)


yoavrotems:
Hey first thank you for contributing. Secondly we had issue with the integration testing, please rebase main and update, when it will pass the test I will approve it.

yoavrotems:
Hey, thanks for contributing, where are those paths from? Could you attach a link supporting those paths? maybe from Talos docs
?

lennardk:
Hi,

I'm not aware of much documentation on them, I've mostly found them by "looking around".
For the `talos-` prefix on manifests, see [these sources lines 238-242)[https://github.com/talos-systems/talos/blob/44a63e9a4d2b26935404e34a474c7df0651fe2ee/pkg/machinery/constants/constants.go#L238].
The same file lines 174+ specify the `/system/secrets` directory, although I've found the paths by simply looking at the command line arguments of the running processes.

yoavrotems:
This is a bit what I was worrying about, since if we want to support it, we need something stable to relay on, since we don't want to add those paths to discover this is not the case for every installation or that in the next minor release it will be changed and then we will "stop" support those.
I think we wouldn't except this if we can't have a leg to stand on... 
I'm tending not to approve this, but what do you think? 
(By the way maybe you could automate taking the command line parameters and check for those files and have it as a flag something like --detect-configs this will be great and more general)

lennardk:
I've reached out to the Talos team and I've received the following response.

> The manifests should be perfectly stable: that directory is expected by the kubelet, and the names are perfectly descriptive and sufficiently differentiated.
The kubelet configs also strike me as unlikely to change.
I'm less certain about the secrets, but even there, they're in a good place with good names, and we have no plans to change them.

Automating something based on the running processes sounds like a good idea but I'm currently unsure how to go about that. Would the response above indicate enough stability for the present time?

yoavrotems (approved):

yoavrotems:
@lennardk OK then lets do it,

PR #1010, opened 2021-10-01
Fix status of cis-1.20 1.2.25 should be Manual

mozillazg:
https://workbench.cisecurity.org/sections/665102/recommendations/1090253

image


mozillazg:
The integration testing is not working as expected and can be fixed via #1011

PR #1012, opened 2021-10-01
update to go 1.17

neha-viswanathan:
Update to go 1.17 and also update go modules


mozillazg:
Thanks for your contribution! And sorry for the delay. We'd like only update `go 1.16` to `go 1.17` and use @dependabot to update go modules like https://github.com/aquasecurity/kube-bench/pull/1076

PR #1018, opened 2021-10-09
use k8s.io/klog/v2 instead of github.com/golang/glog

mmorel-35:


yoavrotems:
I tried to check about glog not maintained but didn't find anything special and I think its currently OK with us, do you have more information about that?

PR #1030, opened 2021-10-23
Add --status flag

### LINKED ISSUE #918: New flag! --status
After [Add new statuses](#916) kube-bench will have 7 different statuses 
[PASS] [FAIL] [WARN] [INFO] [MANU] [SKIP] [ERRO] 
It only make sense to have a new output manipulation flag --status which will get a string array of statuses and will only print test with the desired status for example: 
`kube-bench --status="ERRO"` 
will only output tests with ERRO status 
and 
`kube-bench --status="SKIP,MANU" `
will output the skipped and manual tests.

mtpereira:
Fixes #918 

This change adds the `--status` flag as described on the referenced issue, so that `kube-bench` will only log tests results with a `check.State` contained in the `--status` flag. If the flag is not passed, all messages are logged.
This does not affect the summary output.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -205,24 +205,29 @@ func prettyPrint(r *check.Controls, summary check.Summary) {
 
 	// Print remediations.
 	if !noRemediations {
+		var remediationOutput strings.Builder
 		if summary.Fail > 0 || summary.Warn > 0 {
-			colors[check.WARN].Printf("== Remediations %s ==\n", r.Type)
 			for _, g := range r.Groups {
 				for _, c := range g.Checks {
-					if c.State == check.FAIL {
-						fmt.Printf("%s %s\n", c.ID, c.Remediation)
+					if c.State == check.FAIL && printStatus(check.FAIL) {
+						remediationOutput.WriteString(fmt.Sprintf("%s %s\n", c.ID, c.Remediation))
 					}
-					if c.State == check.WARN {
+					if c.State == check.WARN && printStatus(check.WARN) {
 						// Print the error if test failed due to problem with the audit command
 						if c.Reason != "" && c.Type != "manual" {
-							fmt.Printf("%s audit test did not run: %s\n", c.ID, c.Reason)
+							remediationOutput.WriteString(fmt.Sprintf("%s audit test did not run: %s\n", c.ID, c.Reason))
 						} else {
-							fmt.Printf("%s %s\n", c.ID, c.Remediation)
+							remediationOutput.WriteString(fmt.Sprintf("%s %s\n", c.ID, c.Remediation))
 						}
 					}
 				}
 			}
-			fmt.Println()
+			output := remediationOutput.String()
+			if len(output) > 0 {
+				remediationOutput.WriteString("\n")
+				fmt.Printf(colors[check.WARN].Sprintf("== Remediations %s ==\n", r.Type))
+				fmt.Printf(remediationOutput.String())
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


mtpereira:
To make this more explicit and hopefully easier to communicate, here is the current output I get from integration tests when I modify them to pass `--status=FAIL`: https://gist.github.com/mtpereira/12e0acb97a209b2ae72389d3a640ca28

yoavrotems:
Hey you were right in the first part 


> Hi @yoavrotems ! 👋
> 
> I'm not sure I understood what you meant with your comment on what the output should be, so I'll try to clarify it. 🙏
> 
> Do you mean that:
> 
> 1. when the --status flag is set, all of the stdout should be filtered, including summary and report;
> 2. Or do you mean that other methods of outputting should also be affected, with the exception of JSON, as you've mentioned?
> 
> Thanks for the clarification!

About the second part all of the output manipulation flags --noremediations --nosummary --noresults effects the stdout only and not the json. 
And yes it should all co-exist I don't see a reason why not using --noremediations and --status=FAIL

I saw the output you shared status wise it seems good, but I would skip printing remediation for not printed tests like you have with 3.1.1 3.2.1 3.2.2 those tests all are WARN so it not printed but the remediation do. 

If someone wanted to see all fails tests the it remediation I guess that WARN tests remediation is not relevant here. if so you should run --status="FAIL,WARN"

wdyt?

mtpereira:
@yoavrotems Got it, that makes sense! I'll give it a look. :ok_hand:

mtpereira:
Hello @yoavrotems :wave:

How is this looking now? https://gist.github.com/mtpereira/12e0acb97a209b2ae72389d3a640ca28/revisions

:memo: The code needs some refactoring, it is still not ready, but tests are :green_circle: .

yoavrotems:
Hey looks better the output that not including remediation when its not printing the test is great! 
This PR is great and part from a bigger change that we want to see, I think you should take in consideration this PR #1035

mtpereira:
> Hey looks better the output that not including remediation when its not printing the test is great! 
> 
> This PR is great and part from a bigger change that we want to see, I think you should take in consideration this PR #1035  


Hi! 👋 

Oh this might be tricky to handle! 😅 How do you suggest we go about this? Shall we merge one PR first and then proceed with the other? I'm fine either way. 👍

yoavrotems:
Let me check it and return to you on how we are going to accept the other PR because we might take as a flag to not break old behaviors

yoavrotems:
Hey @mtpereira We would accept the #1035 PR but we will do it as a flag for now, the default will be using all of the statuses, but with flag `--enhancedoutput=false`  to allow backward compatibility, so I think that this PR will need to support filtering all of the statuses. 
One thing we need to decide is the behavior when printing unused flag, for example:
 `./kube-bench --enhancedoutput=false --status="WARN,MANU"` 

I think that we can do in one of the two ways, 
1. When use flag `status` check what filtering is applying and if using new status with `enhancedoutput`=false, raise error and not running. --> `kube-bench can't output non-supported status, either remove $STATUS from --status, or set --enhancedoutput=true` P.S should also support for something like this one trying none exist status.

2. When trying to filter out none exist flag just not print it since there isn't any $STATUS in the output, same as I would run --status="FAIL" and none of the tests failed. 

I think that the first option is the best but open to suggestions. 
WDYT?

mtpereira:
Hello @yoavrotems ! :wave: 

Option 1 seems better to me as well! I'll give it a look once I have some spare time. :bow:

Issue #916, opened 2021-06-21
Add new statuses 


yoavrotems:
Currently Kube-bench output only four statuses
* [PASS] - Test passed 
* [FAIL] - Test scored and got unwanted output
* [WARN] - Test is not scored and failed, or test has unknown command or test command failed to run (wrong parameter, file doesn't exist and etc) or test is a manual test
* [INFO] - Test is either chapter title or marked as skip. 

We want to improve it and use new status, this is what I thought about: 
### New status 

- [**INFO**] 
    **Color**: Cyan
    **Use case**: Only for chapter titles
    **Output**: None
    
- [**PASS**] 
    **Color**: Green
    **Use case**: Passing tests scored and not scored
    **Output**: None, no remediation
    
- [**FAIL**] 
    **Color**: Red
    **Use case**: Failed scored tests
    **Output**: Remediation
    
- [**WARN**] 
    **Color**: Yellow
    **Use case**: Failed not scored tests
    **Output**: Remediation
    
- [**SKIP**] 
    **Color**: Blue
    **Use case**: Skip labeled tests
    **Output**: None, no remediation, skipped test
    
- [**MANU**] 
    **Color**: White
    **Use case**: Manual labeled tests
    **Output**: Start with this is a manual test, then remediation
    
- [**ERRO**] 
    **Color**: Magenta
    **Use case**: Error in tests: non exist command or parameter, command return non 0 exit code
    **Output**: Error message from bash output.

PR #1035, opened 2021-10-31
Introduce new statuses: SKIP, MANU, ERRO

hypnoglow:
This changeset introduces new statuses as proposed in #916

Probably I missed something, so I would be grateful any feedback and help 🙏


cmd/common.go
```diff
@@ -184,14 +186,14 @@ func prettyPrint(r *check.Controls, summary check.Summary) {
...
 			for _, g := range r.Groups {
 				for _, c := range g.Checks {
-					if c.State == check.FAIL {
+					if c.State == check.FAIL || c.State == check.ERRO {
 						fmt.Printf("%s %s\n", c.ID, c.Remediation)
 					}
-					if c.State == check.WARN {
+					if c.State == check.WARN || c.State == check.MANU {
 						// Print the error if test failed due to problem with the audit command
 						if c.Reason != "" && c.Type != "manual" {
```
yoavrotems, line 198: WARN should be with FAIL and here it should be ERRO since Reason was a warn_reason which is why a test returned error, now WARN is just a fail test with no score. and ERRO is the test with reasons like command not exist or invalid parameter. 
I think maybe we should consider change 'warn_reason' in the json output to error_reason or something like this.

cmd/common.go
```diff
@@ -173,7 +173,9 @@ func prettyPrint(r *check.Controls, summary check.Summary) {
 			for _, c := range g.Checks {
 				colorPrint(c.State, fmt.Sprintf("%s %s\n", c.ID, c.Text))
 
-				if includeTestOutput && c.State == check.FAIL && len(c.ActualValue) > 0 {
+				if includeTestOutput &&
+					(c.State == check.FAIL || c.State == check.ERRO) &&
```
yoavrotems, line 177: Also WARN should be printed since starting now WARN is a fail just not scored.

yoavrotems (changes requested):
First of all thanks for contributing ! 
Ok it seems good and nice :) there are a small changes I requested just a bit around the logic and about maybe change warn_reason in the json output but I'm not sure about since it will break backward compatibility but on the other hand we are changing a lot here.

cmd/common.go
```diff
@@ -184,14 +186,14 @@ func prettyPrint(r *check.Controls, summary check.Summary) {
...
 	// Print remediations.
 	if !noRemediations {
-		if summary.Fail > 0 || summary.Warn > 0 {
+		if summary.Fail > 0 || summary.Warn > 0 || summary.Manu > 0 || summary.Erro > 0 {
 			colors[check.WARN].Printf("== Remediations %s ==\n", r.Type)
 			for _, g := range r.Groups {
 				for _, c := range g.Checks {
-					if c.State == check.FAIL {
+					if c.State == check.FAIL || c.State == check.ERRO {
 						fmt.Printf("%s %s\n", c.ID, c.Remediation)
```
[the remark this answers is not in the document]
hypnoglow: @yoavrotems Could you elaborate on this? The error is stored in Reason, is that right?

Do you mean something like this?

```go
if c.State == check.ERRO {
    fmt.Printf("%s audit test error: %s\n", c.ID, c.Reason)
}
```

cmd/common.go
hypnoglow: Did I understand you correctly?

```suggestion
					(c.State == check.FAIL || c.State == check.WARN || c.State == check.ERRO) &&
```

cmd/common.go
hypnoglow: I'm not sure I understood what `warn_reason` is, since there is only `reason` in json output.

hypnoglow:
@yoavrotems I've made a few changes based on your comments, PTAL. Although I'm not sure I got everything correct 🙄

yoavrotems:
Before I'm reviewing those new changes, we talked about it with some kube-bench users, and those changes are breaking the output for automation and other projects using kube-bench so lets do it for now as a flag, for example --enhancedoutput, and then it will be default to true, and if someone needs backward compatibility they could run it as `./kube-bench --enhancedoutput=false `.
It's important to say that this output with more status will be the default, and returning to the old status will have to be specifically chosen.  
It will be part of 0.7.0 and in the next versions we will deprecate the old output once most of our consumers adjust the new output

mozillazg:
> Before I'm reviewing those new changes, we talked about it with some kube-bench users, and those changes are breaking the output for automation and other projects using kube-bench so lets do it for now as a flag, for example --enhancedoutput, and then it will be default to true, and if someone needs backward compatibility they could run it as `./kube-bench --enhancedoutput=false `. It's important to say that this output with more status will be the default, and returning to the old status will have to be specifically chosen. It will be part of 0.7.0 and in the next versions we will deprecate the old output once most of our consumers adjust the new output

@hypnoglow Hi, Are you still working on this?

Issue #984, opened 2021-09-11
EKS benchmark rule not found


sahilchadha1991:
https://github.com/aquasecurity/kube-bench/blob/10ba0adb2d45eb45b9178b51ede45d4dfcae46a0/cfg/eks-1.0/policies.yaml#L196

This rule is not part of eks cis 1.0.0 benchmark

PR #1042, opened 2021-11-07
Update aks-1.0 to match official CIS Azure Kubernetes Service (AKS) Benchmark v1.0.0

mozillazg:
[《CIS Azure Kubernetes Service (AKS) Benchmark v1.0.0》](https://workbench.cisecurity.org/benchmarks/6347)

Fixes https://github.com/aquasecurity/kube-bench/issues/984
Fixes https://github.com/aquasecurity/kube-bench/discussions/968
Fixes https://github.com/aquasecurity/kube-bench/discussions/937


yoavrotems (changes requested):
Thanks for contributing, I know there is an issue with empty remediation section in the CIS but lets maybe try to have there something, I suggested to have a different fields there which are not remediation but explaining the logic with the test. 
Maybe we should consider use "No remediation" instead because this is the most correct one but I'm not sure this is the best option. 
WDYT?

mozillazg:
@yoavrotems 

Yes, having empty remediation isn't a good user experience.

+1 for `No remediation`.
Because the data which is picked from other fields is not 100% suitable for use as remediation, even though it may explain the logic with the test. and the `No remediation` matches the origin cis benchmark.

yoavrotems:
OK then lets go with it, `no remediation` 😸

mozillazg:
@yoavrotems Thanks for your review! Please check again.

yoavrotems (approved):

yoavrotems (approved):
LGTM

PR #1045, opened 2021-11-14
Support CIS Amazon Elastic Kubernetes Service (EKS) Benchmark v1.0.1

mozillazg:
CIS Amazon Elastic Kubernetes Service (EKS) Benchmark v1.0.1:
https://workbench.cisecurity.org/benchmarks/6041


yoavrotems (changes requested):
Start to look at the files all in all looks good just need to check in compare to CIS that its ok. 
But if we are updating to eks 1.0.1 lets do it correctly and update the config.yaml for the version mapping, change the cfg dir name, I see that its not like new spec that talks about version X and above but correct me if I'm wrong its completely replacing the old spec, if it is the case then really lets just rename it to be eks-1.0 but if I'm wrong then maybe lets add it as a new spec and not editing eks-1.0 
P.S 
If it really replacing the eks-1.0 I think that we wouldn't need to change the auto-detection code but its need to be checked.

mozillazg:
The CIS EKS Benchmark v1.0.1 is a [Bug Fix] version, so I just replace the eks-1.0

image

yoavrotems:
Then let change it to be eks-1.0.1 and change the code as well for the auto detection

mozillazg:
> Then let change it to be eks-1.0.1 and change the code as well for the auto detection

Done.

yoavrotems (approved):
LGTM

debu99:
missing 1.0.1 in aquasec/kube-bench:latest image

yoavrotems:
As I replied in #915 
> It's not missing but just not there yet, it will be released in the next release, docker image used to be created from each push to main, but we changed it to be triggered by new tags. Because it used to create unstable images for each new push. If you have any more questions in the subject please open a issue about it. Thanks

PR #1050, opened 2021-11-21
Support CIS Google Kubernetes Engine (GKE) Benchmark v1.2.0

mozillazg:
Closes https://github.com/aquasecurity/kube-bench/discussions/1049


yoavrotems (changes requested):
Thanks for contributing, I think that in this case in difference from what we did in #1045 we need to create new cfg, then we need to separate it by this version is for kubernetes gke 1.20 + and the old one (gke-1.0) is gke 1.15 - 1.19 I don't think it will take much since we are already supporting this in other platforms. 
Beside this, I think that something that we also need to change in the old version is that test that has:
> remediation: "This control cannot be modified in GKE."

https://github.com/aquasecurity/kube-bench/pull/1050/files#diff-cb00954ead0946c1de96599f6b00338f3abec87bf68a05d920f149af32586e05R27

We should use the type: skip, and not manual as we used to (because this is manual from since before having the type:skip) because this is not a manual running test but a test that we want to just skip. 

Beside those notes, I'm checking with the new benchmark that anything else is in order (just the names, ids and audit commands) It will take me around a day or two... 
In the meanwhile I will be happy if you could address those notes.

mozillazg:
@yoavrotems OK. I have restored the `gke-1.0` and use `type: skip` for `remediation: "This control cannot be modified in GKE."`

I don't update changes for the auto-detection code yet, because it is not easy to support:
> gke-1.2.0 for kubernetes gke 1.20 + and the old one (gke-1.0) is gke 1.15 - 1.19

yoavrotems:
OK for now lets keep it that the auto-detection will chose the gke-1.2 as you did, 
and the old version could be manually ran with `--benchmark`

yoavrotems:
LGTM

Issue #1000, opened 2021-09-20
The `run` command does not appear to honor the `--exit-code` flag


andrewmyhre:
**Overview**

The `--exit-code` command, proposed [here](https://github.com/aquasecurity/kube-bench/discussions/566) and implemented [here](https://github.com/aquasecurity/kube-bench/pull/734), seems to _only_ be honored by the `root` command. It seems to be ignored by the `run` command. 

**How did you run kube-bench?**

In a Kubernetes job with the arguments:

```
          image: aquasec/kube-bench:v0.6.4
          command: ["/usr/local/bin/kube-bench"]
          args:
          - "run"
          - "--benchmark=cis-1.6"
          - "--exit-code=1"
```

**What happened?**

Even though a control is failing, the exit code from the pod is still `0`.

**What did you expect to happen:**

I expected the exit code from the pod to be `1` as I specified in the kube-bench arguments.

**Environment** 

kube-bench version: v0.6.4

Kubernetes version: v0.17.3

**Running processes**

I'm fairly sure the problem is that `exitCodeSelection()` is invoked from `root.go` and nowhere else.

andrewmyhre:
Woohoo 1000th issue!

yoavrotems:
I think its relate to issue with the 0.6.4 build not building, I will release a new one today

yoavrotems:
I'm closing this issue, if it is still an issue in 0.6.5 please reopen this.

ispirals:
Hi, I'm having the same issue using the latest image with the `run` subcommand. 
```
 ["kube-bench", "run", "--targets", "node", "--benchmark", "eks-1.0.1", "--exit-code", "42"]
```

mozillazg:
@ispirals Thanks report.

mozillazg:
The PR #1084 will fix this issue.

PR #1084, opened 2022-01-21
Fix the `--exit-code` flag doesn't work when run with subcommand

mozillazg:
Fixes https://github.com/aquasecurity/kube-bench/issues/1000


yoavrotems (approved):
LGTM

PR #1085, opened 2022-01-24
Update kube bench docker image tag to v0.6.6

sdd1798:


job.yaml
```diff
@@ -12,7 +12,7 @@ spec:
       hostPID: true
       containers:
         - name: kube-bench
-          image: aquasec/kube-bench:0.6.3
+          image: aquasec/kube-bench:v0.6.6
```
mozillazg, line 15: How about changing it to `latest` ?

sdd1798: Not sure if `latest` will be always stable. Also best practice is to use specific tag instead of latest.

mozillazg (approved):

PR #1126, opened 2022-03-23
Syntax errors in node targets for OCP

sm171190:
Fixing syntax error in checks 4.1.1,4.1.2 for rhel-1.0.
SLK-39504

### DIFF
--- a/cfg/rh-1.0/node.yaml
+++ b/cfg/rh-1.0/node.yaml
@@ -11,9 +11,9 @@ groups:
       - id: 4.1.1
         text: "Ensure that the kubelet service file permissions are set to 644 or more restrictive (Automated)"
         audit: |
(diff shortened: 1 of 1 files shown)


mozillazg:
@sm171190 Thanks for your contribution.

The audit contents look like don't have a syntax problem.

IMHO, the root cause is the `oc` command was running failed due to miss profile file(https://docs.openshift.com/container-platform/4.6/cli_reference/openshift_cli/managing-cli-profiles.html) in the container. Please change to mount a profile file to the container then try again. @chen-keinan Could you help me verify this guess? Because I don't have an OCP account.

chen-keinan:
@mozillazg might take time to look into it.  

@sm171190 this fix seems to be quite simple, Please confirm it solve the root cause of the issue

PR #1206, opened 2022-06-10
fix: fully qualified image names

rhtenhove:
Confused as to why my CRIO backed cluster wouldn't run the jobs after a bit of testing, I found out it was trying `quay.io/aquasec/kube-bench` because the docker registry failed due to rate limiting.

The current image names assumes everyone defaults to `docker.io` as their registry, but this seems to no longer be a very safe assumption due to the recent changes by Docker on rate limiting.

This PR prevents a container runtime to make wrong assumptions and fail unclearly (i.e. pull unauthorized errors).

A different, but still relatively short term, solution would be to also publish this image to `quay.io` as that is where the cool kids seem to be moving to (no rate limiting!). But then you'd also need to add `ghcr.io`, and `gcr.io`, and ...

P.S.: my VS Code plugins also did some formatting I see. They do no harm and increase readability, but can be removed if so desired.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/aquasecurity/kube-bench
 go 1.16
 
 require (
-	github.com/aws/aws-sdk-go v1.44.16
+	github.com/aws/aws-sdk-go v1.44.31
 	github.com/fatih/color v1.13.0
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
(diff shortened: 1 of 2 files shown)


mozillazg (approved):

rhtenhove:
> @rhtenhove thanks you for the contribution, could you please rebase your branch with upstream

done :)

PR #1246, opened 2022-08-16
fix kube-bench/cfg/cis-1.6/master.yaml 1.2.26

jonipatr:
Add test to 1.2.26 to check that --request-timeout parameter is configured.

### DIFF
--- a/cfg/cis-1.6/master.yaml
+++ b/cfg/cis-1.6/master.yaml
@@ -714,7 +714,9 @@ groups:
       - id: 1.2.26
         text: "Ensure that the --request-timeout argument is set as appropriate (Automated)"
         audit: "/bin/ps -ef | grep $apiserverbin | grep -v grep"
-        type: manual
+        tests:
+          test_items:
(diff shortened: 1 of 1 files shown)


cfg/cis-1.6/master.yaml
```diff
@@ -714,7 +714,9 @@ groups:
       - id: 1.2.26
         text: "Ensure that the --request-timeout argument is set as appropriate (Automated)"
         audit: "/bin/ps -ef | grep $apiserverbin | grep -v grep"
-        type: manual
```
mozillazg, line 717: Marking it as manual is expected according to the description in cis 1.6: 
image

mozillazg: Here is the origin PR which change it to manual: https://github.com/aquasecurity/kube-bench/pull/874

afdesk:
based on https://github.com/aquasecurity/kube-bench/pull/1246#discussion_r950813951 we can close this PR

PR #1260, opened 2022-08-24
Add label to eks jobs

martelo:
Adding label to eks jobs so we can inspect results using `kubectl logs --selector=app=kube-bench`

### DIFF
--- a/job-eks-asff.yaml
+++ b/job-eks-asff.yaml
@@ -26,6 +26,9 @@ metadata:
   name: kube-bench
 spec:
   template:
+    metadata:
+      labels:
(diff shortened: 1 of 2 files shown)


mozillazg:
@martelo Thanks for your contribution! Looks like we can inspect results using `kubectl logs -l job-name=kube-bench` without adding a new label.

martelo:
@mozillazg fair enough, do you want to close without merging?

mozillazg:
> @mozillazg fair enough, do you want to close without merging?

@martelo Yes, I would prefer to close without merging.

PR #1261, opened 2022-08-26
feat(cis-1.6-k3s): Add support to CIS-1.6 for k3s distribution

TARI0510:
kube-bench is mainly check security benchmark in k8s. In [k3s](https://k3s.io/) distribution, there are many difference from it, for example, 
someone in issue https://github.com/k3s-io/k3s/issues/826 needs a validation for k3s distribution.

So I develop a k3s cis-1.6 version according to k3s security document
+ https://docs.rancher.cn/docs/k3s/security/self-assessment/_index
+ https://rancher.com/docs/k3s/latest/en/security/self_assessment/
+ https://rancher.com/docs/k3s/latest/en/security/hardening_guide/

[Test Env]
```
[root@k3s db]# cat /etc/redhat-release
CentOS Linux release 7.9.2009 (Core)
[root@k3s db]# k3s -v
k3s version v1.24.3+k3s1 (990ba0e8)
go version go1.18.1
```

[build]
```bash
go build github.com/aquasecurity/kube-bench/
```

[run]
```bash
kube_bench run --benchmark cis-1.6-k3s
```

[result]
Please see in https://pastebin.com/dSg7xHT9


mozillazg:
@TARI0510 Thanks for your contribution! I will take a look at it later.

cfg/cis-1.6-k3s/controlplane.yaml
```diff
@@ -0,0 +1,40 @@
+---
+controls:
+version: "cis-1.6"
```
mozillazg, line 3:
```suggestion
version: "cis-1.6-k3s"
```

cfg/cis-1.6-k3s/etcd.yaml
```diff
@@ -0,0 +1,129 @@
+---
+controls:
+version: "cis-1.6"
```
mozillazg, line 3: ditto.

cfg/cis-1.6-k3s/master.yaml
```diff
@@ -0,0 +1,872 @@
+---
+controls:
+version: "cis-1.6"
```
mozillazg, line 3: ditto.

cfg/cis-1.6-k3s/node.yaml
```diff
@@ -0,0 +1,269 @@
+---
+controls:
+version: "cis-1.6"
```
mozillazg, line 3: ditto.

cfg/cis-1.6-k3s/policies.yaml
```diff
@@ -0,0 +1,260 @@
+---
+controls:
+version: "cis-1.6"
```
mozillazg, line 3: ditto.

mozillazg:
@TARI0510 Would you please update the document at https://github.com/aquasecurity/kube-bench/blob/main/docs/platforms.md to include this new benchmark?

mozillazg (changes requested):

TARI0510:
> @TARI0510 Would you please update the document at https://github.com/aquasecurity/kube-bench/blob/main/docs/platforms.md to include this new benchmark?

Of course

TARI0510:
@mozillazg I have modified it

ChrisHolman:
What's the status of this? Very keen to try this out on K3s!

cfg/cis-1.6-k3s/master.yaml
```diff
@@ -0,0 +1,872 @@
...
+                value: "644"
+        remediation: |
+          Run the below command (based on the file location on your system) on the master node.
+          For example,
+          chmod -R 644 /var/lib/rancher/k3s/server/tls/*.crt
+        scored: true
+
+      - id: 1.1.21
+        text: "Ensure that the Kubernetes PKI key file permissions are set to 600 (Automated)"
+        audit: "find //var/lib/rancher/k3s/server/tls/ -name '*.key' | xargs stat -c permissions=%a"
```
mozillazg, line 189:
```suggestion
        audit: "find /var/lib/rancher/k3s/server/tls/ -name '*.key' | xargs stat -c permissions=%a"
```

cfg/cis-1.6-k3s/master.yaml
```diff
@@ -0,0 +1,872 @@
...
+          chmod -R 600 /etc/kubernetes/pki/*.key
+        scored: true
+
+  - id: 1.2
+    text: "API Server"
+    checks:
+      - id: 1.2.1
+        text: "Ensure that the --anonymous-auth argument is set to false (Automated)"
+        audit: |
+          res=$(journalctl -u k3s | grep "Running kube-apiserver" | tail -n1 | grep "anonymous-auth");
```
mozillazg, line 209: It looks like this command can't work as expected when running in a container, but the result will be passed.

BTW, All checks that use this way to audit have the same problem.

cfg/cis-1.6-k3s/master.yaml
```diff
@@ -0,0 +1,872 @@
...
+        text: "Ensure that the Container Network Interface file permissions are set to 644 or more restrictive (Not Applicable)"
+        scored: false
+
+      - id: 1.1.10
+        text: "Ensure that the Container Network Interface file ownership is set to root:root (Not Applicable)"
+        scored: false
+
+      - id: 1.1.11
+        text: "Ensure that the etcd data directory permissions are set to 700 or more restrictive if etcd is used (Automated)"
+        audit: if [ -d "/var/lib/rancher/k3s/server/db/etcd" ];then stat -c permissions=%a /var/lib/rancher/k3s/server/db/etcd; else echo 'NotEtcd'; fi
```
mozillazg, line 53: > if etcd is used
> 'NotEtcd'

I can't find the source from k3s document for this case.

mozillazg (changes requested):

mozillazg:
@TARI0510 Thanks for your contribution! Added some comments please check again, thanks!

cfg/cis-1.6-k3s/master.yaml
TARI0510: It's from https://docs.rancher.cn/docs/k3s/security/self-assessment/_index/ of 1.1.11. 

> if etcd is used 'NotEtcd'

This is my idea, If the path `/var/lib/rancher/k3s/server/db/etcd` not found, take it pass to prevent false positives, because not the test case false, it's only the file path cannot be found.

cfg/cis-1.6-k3s/master.yaml
TARI0510: It may be the container doesn't have `journalctl` command. Should I add the judgment of the other case while `journalctl` command not found?

TARI0510: I wrote this to reduce the false positive rate of the tool, forgetting to consider the case when the command cannot be found

TARI0510:
> @TARI0510 Thanks for your contribution! Added some comments please check again, thanks!

OK, I will check it later

cfg/cis-1.6-k3s/master.yaml
mozillazg: What if the k3s server use another directory as etcd data directory? In this case, it should fail instead of pass.
IMHO, we should not allow the check to be passed when the file path cannot be found.

cfg/cis-1.6-k3s/master.yaml
mozillazg: > It may be the container doesn't have journalctl command. Should I add the judgment of the other case while journalctl command not found?

IMHO, when the `journalctl -u k3s | grep "Running kube-apiserver"` command failed, the check should fail. Not only because of `journalctl` command may not be found but also because the running logs may not be found too.

mozillazg (changes requested):

TARI0510:
@mozillazg Hey, I correct the problem you mention above, Please have a look.

mozillazg:
@TARI0510 Thanks for your patience. Would you please run the kube-bench command again and then upload the results? Thanks!

```
kube_bench run --benchmark cis-1.6-k3s -v 3
```

TARI0510:
@mozillazg this is the result: https://pastebin.com/92nU6059

mozillazg:
> The E2e tests build failed, which seems is not my code change lead to

@TARI0510 The e2e tests will be fixed after https://github.com/aquasecurity/kube-bench/pull/1278 is merged

mozillazg (approved):

mozillazg:
@TARI0510 Would you please rebase your branch with upstream, Thanks!

TARI0510:
> @TARI0510 Would you please rebase your branch with upstream, Thanks!

Yeap~

chen-keinan (approved):

Issue #1275, opened 2022-09-13
ETCD ownership related checks fail with job.yaml


i-karina:
**Overview**

ETCD ownership related checks will always failed when running in container. 
Issue similar to https://github.com/aquasecurity/kube-bench/issues/842 which related to job-master.yaml
Mounting directories from host in job.yaml helps with the issue: `/erc/passwd` and `/etc/group`

**How did you run kube-bench?**

`kubectl apply -f job.yaml `

**What happened?**

ETCD ownership check always fail:

`$ stat -c %U:%G /var/lib/etcd
etcd:etcd`

`$ sudo ls -la /var/lib/ | grep etcd
drwx------  3 etcd           etcd           4096 Sep 12 13:43 etcd
`

`
[FAIL] 1.1.12 Ensure that the etcd data directory ownership is set to etcd:etcd (Automated)
...
1.1.12 On the etcd server node, get the etcd data directory, passed as an argument --data-dir,
from the command 'ps -ef | grep etcd'.
Run the below command (based on the etcd data directory found above).
For example, chown etcd:etcd /var/lib/etcd
`

` ps -ef | grep etcd
root       10346   10256  2 Sep12 ?        00:26:53 etcd --advertise-client-urls=https://192.168.122.109:2379 --cert-file=/etc/kubernetes/pki/etcd/server.crt --client-cert-auth=true --data-dir=/var/lib/etcd `

**What did you expect to happen:**

1.1.12 Ensure that the etcd data directory ownership is set to etcd:etcd (Automated)
check passes

**Environment** 

kube-bench run from main branch:
https://github.com/aquasecurity/kube-bench/tree/main
(image: docker.io/aquasec/kube-bench:v0.6.9)

K8s 1.23.3

PR #868, opened 2021-05-05
mount /etc/passwd and /etc/group for etcd ownership related checks

mozillazg:
Fixes #842

Issue #842, opened 2021-03-07
ETCD ownership related checks will always failed when running in container


mozillazg:
**Overview**

ETCD ownership related checks will always failed when running in container. Because of there is no user named `etcd` in container.

Maybe we should mount the `/etc/passwd` and `/etc/group` from host when running in container (need update README.md and job-master.yaml)?

**How did you run kube-bench?**

```
kubectl apply -f job-master.yaml 
```

**What happened?**

ETCD ownership related checks always failed.

https://github.com/aquasecurity/kube-bench/blob/50fce51da7f231b02668af4a4bb34bf6dbdf0e04/cfg/cis-1.6/master.yaml#L158-L164

**What did you expect to happen:**

Not failed when the ownership of ETCD data-dir is `etcd:etcd`.

**Environment** 

[What is your version of kube-bench? (run `kube-bench version`)]

v0.5.0

yoavrotems:
Hey @mozillazg thanks for informing us about this issue, currently I'm working on different issues but I will mark it as good first issue and help wanted

PR #1276, opened 2022-09-13
mount /etc/passwd and /etc/group for etcd ownership related checks

i-karina:
Changes to job-master.yaml are also needed in job.yaml to properly check ownership of etcd data dir

Issue: https://github.com/aquasecurity/kube-bench/issues/1275

Related to issue: https://github.com/aquasecurity/kube-bench/issues/842
fix the same as in https://github.com/aquasecurity/kube-bench/pull/868

### DIFF
--- a/job.yaml
+++ b/job.yaml
@@ -50,6 +50,12 @@ spec:
             - name: opt-cni-bin
               mountPath: /opt/cni/bin/
               readOnly: true
+            - name: etc-passwd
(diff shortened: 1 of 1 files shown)


mozillazg:
In most cases the job that was created via job.yaml will not be running on the master node. 
IMHO, mount /etc/passwd and /etc/group from worker node is not necessary.

Issue #1439, opened 2023-05-12
Kube-bench Fail to run on Tencent Kubernetes Engine Orchestrator


deven0t:
**Overview**

Whenever we are running kube-bench on Tencent Kubernetes Engine Orchestrator(Tencent Cloud ) cluster, it exists with an error 
`error looking for file /etc/kubernetes/kubelet/kubelet-config.json: stat /etc/kubernetes/kubelet/kubelet-config.json: not a directory`

**How did you run kube-bench?**

We run with pod yaml file. It is same as this one https://github.com/aquasecurity/kube-bench/blob/main/job-node.yaml

**What happened?**
Pod gives error in the log
`error looking for file /etc/kubernetes/kubelet/kubelet-config.json: stat /etc/kubernetes/kubelet/kubelet-config.json: not a directory`

```
➜ ✗ kubectl logs kube-bench-4td6d 

error looking for file /etc/kubernetes/kubelet/kubelet-config.json: stat /etc/kubernetes/kubelet/kubelet-config.json: not a directory
```

mozillazg:
Fixed via https://github.com/aquasecurity/kube-bench/pull/1440

PR #1440, opened 2023-05-12
fix: ignore the error from findConfigFile

deven0t:
When we are trying to access a file from a directory which is not present then we get different error. We dont have standard error method to check the msg so added string match for this case
#1439


deven0t:
Hi @chen-keinan 
Can you review this?

Thanks,
Devendra

chen-keinan:
@deven0t lgtm , just need rebase

chen-keinan (approved):

PR #1513, opened 2023-10-26
Add CIS Benchmarks support to Rancher Distributions RKE/RKE2/K3s

KiranBodipi:
Based on the information furnished in https://ranchermanager.docs.rancher.com/v2.7/pages-for-subheaders/rancher-hardening-guides#hardening-guides-and-benchmark-versions , kube-bench executes CIS-1.23 (Kubernetes v1.23) , CIS-1.24(Kubernetes v1.24),CIS-1.7 (Kubernetes v1.25,v1.26,v1.27) CIS Benchmarks of respective distributions.

PR #1523, opened 2023-11-16
Add CIS Benchmarks support to Rancher Distributions RKE/RKE2/K3s

KiranBodipi:
Based on the information furnished in https://ranchermanager.docs.rancher.com/v2.7/pages-for-subheaders/rancher-hardening-guides#hardening-guides-and-benchmark-versions , kube-bench executes CIS-1.23 (Kubernetes v1.23) , CIS-1.24(Kubernetes v1.24),CIS-1.7 (Kubernetes v1.25,v1.26,v1.27) CIS Benchmarks of respective distributions.
Benchmark files are taken from here https://github.com/rancher/security-scan/tree/master/package/cfg


deven0t (approved):

KiranBodipi:
Hey @mozillazg @chen-keinan  could you kindly assist in reviewing current PR instead of https://github.com/aquasecurity/kube-bench/pull/1513, as it was accidentally closed. Thanks.

cfg/config.yaml
```diff
@@ -28,6 +28,7 @@ master:
       - "apiserver"
       - "openshift start master api"
       - "hypershift openshift-kube-apiserver"
+      - "containerd"
```
mozillazg, line 31: IMHO, we should refrain from adding `containerd` to every configuration in this file. For this specific case, the configuration should be specified in each benchmark's dedicated configuration file, such as `cfg/k3s-cis-1.23/config.yaml`.

cmd/util.go
```diff
@@ -290,13 +295,34 @@ Alternatively, you can specify the version with --version
...
+		glog.V(3).Infof("Error fetching cluster config: %s", err)
+	}
+	isRKE := false
+	if err == nil && kubeConfig != nil {
+		k8sClient, err := kubernetes.NewForConfig(kubeConfig)
+		if err != nil {
+			glog.V(3).Infof("Failed to fetch k8sClient object from kube config : %s", err)
+		}
+		if err == nil {
+			isRKE, err = providers.IsRKE(context.Background(), k8sClient)
```
mozillazg, line 309: Can't we determine if the cluster is an RKE cluster based on the version of the apiserver?
If that's the case, I personally suggest implementing this simple `IsRKE` function in the kube-bench project, rather than introducing a dependency on https://github.com/rancher/kubernetes-provider-detector .

mozillazg (changes requested):
Thanks for your contribution! I've added some comments. Please check them when you get a chance. Thanks!

mozillazg:
@KiranBodipi Could you please resolve the conflict? Thanks!

mozillazg (approved):

chen-keinan (approved):

neu7ron2:
The k3s checks don't work properly (if you have used a stand alone K3s instance), for instance
1.1.11 audit check is trying to call some custom script file "check_for_k3s_etcd.sh 1.1.11" which doesn't exist.
Also other check like "journalctl -D /var/log/journal  -u k3s  | grep 'Running kube-apiserver' " takes minutes to complete so should be cached once and reused.

PR #1546, opened 2024-01-10
chore: update go version to 1.21

deven0t:
Updating go version to 1.21


deven0t:
Thank @chen-keinan  for reviewing, I have addressed the comment.

PR #1597, opened 2024-04-12
perf: Improve KubeBench Performance on OCP Cluster

mjshastha:
Currently, certain commands involve retrieving all node names or pods and then executing additional commands in a loop, resulting in a time complexity linearly proportional to the number of nodes.This approach becomes time-consuming for larger clusters.

As kube-bench is executed as a job on every node in the cluster, To enhance performance, Streamlined the commands to execute directly on current node where kube-bench operates. This change ensures that the time complexity remains constant, regardless of the cluster size. By running the necessary commands only once per node, regardless of how many nodes are in the cluster, this approach significantly boosts performance and efficiency.

### DIFF
--- a/cfg/rh-1.0/etcd.yaml
+++ b/cfg/rh-1.0/etcd.yaml
@@ -11,16 +11,17 @@ groups:
       - id: 2.1
         text: "Ensure that the --cert-file and --key-file arguments are set as appropriate (Manual)"
         audit: |
-          # For --cert-file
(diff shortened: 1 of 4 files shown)


deven0t (approved):
LGTM

KiranBodipi (approved):
LGTM

mjshastha:
@chen-keinan Could you verify and merge ?

chen-keinan (approved):

PR #1614, opened 2024-05-15
Align correct cis benchmarks with k8s versions

dereknola:
Per offical CIS benchmark PDFs, the cis versions are slightly off in the docs

CIS v1.7.1 calls out
> This document provides prescriptive guidance for establishing a secure configuration
> posture for Kubernetes v1.25

And cis v1.8.0 calls out
> This document provides prescriptive guidance for establishing a secure configuration
> posture for Kubernetes v1.27.

Since there is no mention of v1.26, I took the assumption the CIS v1.7.1 covers it. 

FYI: cis 1.9 whenever that gets done in kube-bench is supposed to:
>This document provides prescriptive guidance for establishing a secure configuration
> posture for Kubernetes v1.27 - v1.29.

Signed-off-by: Derek Nola <derek.nola@suse.com>


andypitcher (commented):
@dereknola CIS-1.7.1 was a bug release, so it kept the same targeted versions as CIS-1.7 (k8s 1.25).

Here is the K8s version scope for each CIS that was defined by CIS Workbench (strangely CIS-1.9 and 1.10 are overlapping, but maybe that would be fixed once 1.10 is released):

- [CIS-1.7](https://workbench.cisecurity.org/benchmarks/11107) and [CIS-1.7.1](https://workbench.cisecurity.org/benchmarks/13161) : K8s 1.25
- [CIS-1.8](https://workbench.cisecurity.org/benchmarks/12958): K8s 1.26
- [CIS-1.9](https://workbench.cisecurity.org/benchmarks/16828): K8s 1.27-1.29
- [CIS-1.10](https://workbench.cisecurity.org/benchmarks/17568) (draft): k8s 1.28-1.30

dereknola:
Its quite annoying that there is conflicting information between the PDF versions of the CIS benchmark and the online version of the benchmark :angry:

PR #1679, opened 2024-09-16
FIXING RKE CIS CHECKS

sm171190:
In this change we are making 2 changes:
1. adding the check Type as manual for some manual checks for which the type was missing or incorrect
2. Some checks are removed as they are not applicable for an RKE2 cluster
3. checks 1.3.6 and 4.2.12 chekc whether the cluster has been configured with the flags to rotate TLS certificates. However, RKE Internally handles certificate rotation : https://github.com/rancher/dashboard/issues/4485


cfg/ack-1.0/master.yaml
```diff
@@ -896,25 +896,7 @@ groups:
           --root-ca-file=<path/to/file>
         scored: true
 
-      - id: 1.3.6
```
mjshastha, line 899: Rather than removing the check and disrupting the sequence, which may confuse the user, remove 'scored' field and set the type to "skip." If it is set to "skip," it will be marked as "INFO.".

cfg/cis-1.5/node.yaml
```diff
@@ -433,32 +433,6 @@ groups:
           systemctl daemon-reload
           systemctl restart kubelet.service
         scored: true
-
-      - id: 4.2.12
-        text: "Ensure that the RotateKubeletServerCertificate argument is set to true (Scored)"
```
mjshastha, line 438: also here.

cfg/rke-cis-1.24/master.yaml
```diff
@@ -143,10 +143,12 @@ groups:
         audit: |
           ps -ef | grep $kubeletbin | grep -- --cni-conf-dir | sed 's%.*cni-conf-dir[= ]\([^ ]*\).*%\1%' | xargs -I{} find {} -mindepth 1 | xargs --no-run-if-empty stat -c %U:%G
           find /var/lib/cni/networks -type f 2> /dev/null | xargs --no-run-if-empty stat -c %U:%G
+
         use_multiple_values: true
         tests:
           test_items:
             - flag: "root:root"
+        type: manual
```
mjshastha, line 151: What's the rational behind setting the type as "manual" for an automated check? Ideally, we should avoid doing this.

cfg/rke-cis-1.24/master.yaml
```diff
@@ -298,6 +300,7 @@ groups:
               compare:
                 op: bitmask
                 value: "600"
+        type: manual
```
mjshastha, line 303: also here.

What's the reasoning behind setting the type as "manual" for an automated check? Ideally, we should avoid doing this.

cfg/ack-1.0/master.yaml
sm171190: Thanks. Taken care of in new PR - https://github.com/aquasecurity/kube-bench/pull/1685

cfg/rke-cis-1.24/master.yaml
sm171190: This was by mistake. The check number I intended to change in was in a different folder. Correct PR here - https://github.com/aquasecurity/kube-bench/pull/1685

cfg/rke-cis-1.24/master.yaml
sm171190: Same response as above

cfg/cis-1.5/node.yaml
sm171190: Thanks. Tken care of here - https://github.com/aquasecurity/kube-bench/pull/1685

PR #1686, opened 2024-09-19
FIXING RKE2-CIS-1.24 CHECKS

sm171190:
. **MASTER**:
            a. Checks 1.1.10,1.1.20 are manual according to https://docs.rke2.io/security/cis_self_assessment124#1110-ensure-that-the-container-network-interface-file-ownership-is-set-to-root-manual and https://docs.rke2.io/security/cis_self_assessment124#1110-ensure-that-the-container-network-interface-file-ownership-is-set-to-root-manual respectively.
            b. Check 1.3.6 is not relevant to an RKE2 cluster as RKE2 rotates TLS certificates internally - https://github.com/rancher/dashboard/issues/4485. We will skip it and not score it
    2. **NODE**:
            a. Check 4.2.12 is the node-level equivalent of the master-level check 1.3.6 and is treated the same way.
@afdesk  request review please.

### DIFF
--- a/cfg/rke2-cis-1.24/master.yaml
+++ b/cfg/rke2-cis-1.24/master.yaml
@@ -154,6 +154,7 @@ groups:
         tests:
           test_items:
             - flag: "root:root"
+        type: manual
         remediation: |
           Run the below command (based on the file location on your system) on the control plane node.
           For example,
@@ -313,6 +314,7 @@ groups:
                 op: bitmask
                 value: "600"
               set: true
+        type: manual
         remediation: |
           Run the below command (based on the file location on your system) on the control plane node.
           For example,
@@ -979,7 +981,7 @@ groups:
           Edit the Controller Manager pod specification file $controllermanagerconf
           on the control plane node and set the --feature-gates parameter to include RotateKubeletServerCertificate=true.
           --feature-gates=RotateKubeletServerCertificate=true
-        scored: true
+        scored: false
         type: skip
 
       - id: 1.3.7
--- a/cfg/rke2-cis-1.24/node.yaml
+++ b/cfg/rke2-cis-1.24/node.yaml
@@ -440,6 +440,7 @@ groups:
           systemctl daemon-reload
           systemctl restart kubelet.service
         scored: false
+        type: skip
 
       - id: 4.2.13
         text: "Ensure that the Kubelet only makes use of Strong Cryptographic Ciphers (Manual)"


afdesk:
>> Contributor License Agreement is not signed yet.

cfg/rke2-cis-1.24/master.yaml
```diff
@@ -979,7 +981,7 @@ groups:
           Edit the Controller Manager pod specification file $controllermanagerconf
           on the control plane node and set the --feature-gates parameter to include RotateKubeletServerCertificate=true.
           --feature-gates=RotateKubeletServerCertificate=true
-        scored: true
+        scored: false
```
afdesk, line 984: I'm still not sure about this. but if you're - we can merge it

my concerns are next:
* the mentioned issue - https://github.com/rancher/dashboard/issues/4485, added a new feature - Rotate Certificates, if i understood correctly. (last updated on March,2022)
* CIS from RKE2 tells us to [Ensure that the RotateKubeletServerCertificate argument is set to true](https://docs.rke2.io/security/cis_self_assessment124#136-ensure-that-the-rotatekubeletservercertificate-argument-is-set-to-true-automated) (Last updated on Sep 18, 2024)

but we want to ignore it. 

What did I miss? thanks!

sm171190:
> > > Contributor License Agreement is not signed yet.

Have signed it already :)

afdesk:
> > > > Contributor License Agreement is not signed yet.
> 
> Have signed it already :)

Github still doesn't detect it. Could you recheck?

cfg/rke2-cis-1.24/master.yaml
mjshastha: These are two separate features, and based on this, As I see it we should not be skipping the check.

@sm171190 The --feature-gates option has been removed. You can verify this at https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates-removed/ to see if it helps clarify the issue.

cfg/rke2-cis-1.24/node.yaml
```diff
@@ -440,6 +440,7 @@ groups:
           systemctl daemon-reload
           systemctl restart kubelet.service
         scored: false
+        type: skip
```
mjshastha, line 443: also here.

The --feature-gates option has been removed. You can verify this at https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates-removed/ to see if it helps clarify the issue.

cfg/rke2-cis-1.24/master.yaml
```diff
@@ -154,6 +154,7 @@ groups:
         tests:
           test_items:
             - flag: "root:root"
+        type: manual
```
mjshastha, line 157: From the provided link, the check is labeled "manual." What’s the rationale behind setting the type to "manual"? 

We shouldn't classify checks titled "manual" as type "manual," as they have different meanings.

The "manual" type indicates tests that cannot be automated. These tests always yield a WARN state and require user intervention.

CIS’s tool, CAT, has shifted from "score/not scored" to "Automated/Manual" in the 1.6 CIS benchmark to align with its functionality. We follow CIS benchmarks, which is why we use these titles.

cfg/rke2-cis-1.24/master.yaml
```diff
@@ -313,6 +314,7 @@ groups:
                 op: bitmask
                 value: "600"
               set: true
+        type: manual
```
mjshastha, line 317: also here.

From the provided link, the check is labeled "manual." What’s the rationale behind setting the type to "manual"?

We shouldn't classify checks titled "manual" as type "manual," as they have different meanings.

The "manual" type indicates tests that cannot be automated. These tests always yield a WARN state and require user intervention.

CIS’s tool, CAT, has shifted from "score/not scored" to "Automated/Manual" in the 1.6 CIS benchmark to align with its functionality. We follow CIS benchmarks, which is why we use these titles.

cfg/rke2-cis-1.24/master.yaml
sm171190: @mjshastha That is precisely why we want to keep type as manual. Someone using Kube-Bench recently complained about false positives regarding checks that CIS mentioned as Manual. The issue was that these users were seeing a PASS?FAIL fo rthis check and wanted to cross-check. Setting type as manual will mark this check as WARN and not lead to someone thinking that Kube-Bench has incorrectly marked an audit as PAss/FAIL. WARN will, in some sense, force them to run the audit manually to verify - which is the very purpose of a Manual test to begin with ;)

cfg/rke2-cis-1.24/master.yaml
sm171190: @mjshastha That is precisely why we want to keep type as manual. Someone using Kube-Bench recently complained about false positives regarding checks that CIS mentioned as Manual. The issue was that these users were seeing a PASS?FAIL fo rthis check and wanted to cross-check. Setting type as manual will mark this check as WARN and not lead to someone thinking that Kube-Bench has incorrectly marked an audit as PAss/FAIL. WARN will, in some sense, force them to run the audit manually to verify - which is the very purpose of a Manual test to begin with ;)

cfg/rke2-cis-1.24/master.yaml
sm171190: @afdesk Based on the reference shared by @ManojShastha https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates-removed/ the RotateKubeletServerCertificate feature gate was in GA from v1.19 to v1.21. SInce we're dealing with checks on k8s 1.24 I think we can skip it. Whatsay?

cfg/rke2-cis-1.24/master.yaml
mjshastha: If users have previously complained about false positives, it would be more beneficial to verify the check and adjust its implementation accordingly, rather than marking it as type: manual.

Marking it as manual does not address the root cause of the issue and may lead to confusion about the check's validity. We should only set the type to manual if it is truly a non-automatable check.

cfg/rke2-cis-1.24/master.yaml
mjshastha: If users have previously complained about false positives, it would be more beneficial to verify the check and adjust its implementation accordingly, rather than marking it as type: manual.

Marking it as manual does not address the root cause of the issue and may lead to confusion about the check's validity. We should only set the type to manual if it is truly a non-automatable check.

afdesk:
HI guys!
@sm171190 @mjshastha could you confirm the status of this PR?
thanks

andypitcher:
Adding @dereknola for review, the RKE2/K3s profiles have been recently improved to be more accurate.

PR #1688, opened 2024-09-23
FIXING RKE-2-CIS-1.24 Checks

sm171190:
**MASTER**:
            a. Checks **1.1.10,1.1.2**: In The case of manual checks, ideally  the user also checks if the required configuration  files are present or not. We have augmented those checks to capture that information as well, and run the actual audit only when th corresponding configuration files are found.
            b. **Check 1.3.6**: is not relevant to an RKE2 cluster as RKE2 rotates TLS certificates internally - rancher/dashboard#4485. We will skip it and not score it

    2. **NODE**:
            a. Check 4.2.12 is the node-level equivalent of the master-level check 1.3.6 and is treated the same way.

### DIFF
--- a/cfg/rke2-cis-1.24/master.yaml
+++ b/cfg/rke2-cis-1.24/master.yaml
@@ -148,12 +148,18 @@ groups:
       - id: 1.1.10
         text: "Ensure that the Container Network Interface file ownership is set to root:root (Manual)"
         audit: |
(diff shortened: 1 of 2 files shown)


mjshastha (approved):
I'm content with the changes.

sm171190:
> I'm content with the changes.



> I'm content with the changes.

Thanks @ManojShastha . @afdesk Can we proceed with merge?

afdesk (approved):
LGTM! 

@sm171190 could you update the PR description for future?
it seems 1.1.20 and 1.3.6 aren't relevant, but 1.1.21 is missed. thanks!

sm171190:
> LGTM!
> 
> @sm171190 could you update the PR description for future? it seems 1.1.20 and 1.3.6 aren't relevant, but 1.1.21 is missed. thanks!



> LGTM!
> 
> @sm171190 could you update the PR description for future? it seems 1.1.20 and 1.3.6 aren't relevant, but 1.1.21 is missed. thanks!

Good catch.Done!

PR #1722, opened 2024-11-05
FIX| RKE-CIS-1.24- CHECK 1.1.19

sm171190:
The check [1.1.19 of rke-cis-1.24](https://github.com/aquasecurity/kube-bench/blob/main/cfg/rke-cis-1.24/master.yaml#L273-L275) benchmarks relies on a helper audit script that is not present in the Kube-Bench file system. As a result this check results in a file not found error.

nd made it available to the kube-bench file system(https://github.com/rancher/security-scan/blob/master/package/helper_scripts/check_files_owner_in_dir.sh)


cfg/rke-cis-1.24/master.yaml
```diff
@@ -272,7 +272,26 @@ groups:
 
       - id: 1.1.19
         text: "Ensure that the Kubernetes PKI directory and file ownership is set to root:root (Automated)"
-        audit: "check_files_owner_in_dir.sh /node/etc/kubernetes/ssl"
```
deven0t, line 275: we 3 code places where I see this script usage. 
can we try making it available as script and copy them in docker image.

Dockerfile
```diff
@@ -57,3 +59,5 @@ LABEL org.label-schema.build-date=$BUILD_DATE \
     org.label-schema.vcs-ref=$VCS_REF \
     org.label-schema.vcs-url="https://github.com/aquasecurity/kube-bench" \
     org.label-schema.schema-version="1.0"
+
+COPY helper_scripts/check_files_owner_in_dir.sh /go/bin
```
deven0t, line 63: better to have this step before we define entrypoint for better readability. we can put next to other copy commands

deven0t (approved):

afdesk:
hi @sm171190 @deven0t 

I can see this mistake in [the check 1.1.19](https://github.com/aquasecurity/kube-bench/blob/main/cfg/rke-cis-1.24/master.yaml#L273-L275).

helper_scripts/check_files_owner_in_dir.sh
afdesk, line 1: why do we need to copy/paste this script? 
maybe does it make sense to download the script directly from the source?
wdyt?

afdesk (changes requested):

afdesk:
just to confirm, Do i understand correctly, `Dockerfile.ubi` and `Dockerfile.fips.ubi` aren't updated, because there is no any case for RKE CIS and these images, right? thanks!

helper_scripts/check_files_owner_in_dir.sh
sm171190: I think it's better that the script goes packaged with Kube-Bench. wherever one goes, the other csn be guaranteed to be available. And if there are users whose cluster is in an air-gapped env or in a vpc with stringent inbound rules downloading may not even be an option. But this way we are still guaranteed to have the script available. Whatsay? Plu, aesthetically speaking - it improves the structure of Kube-Bench too I feel. Later, we can add our own cusom scripts that replace rpeated actions from multiple audits of other checks. That's my 2 cents. @deven0t ?

sm171190:
> just to confirm, Do i understand correctly, `Dockerfile.ubi` and `Dockerfile.fips.ubi` aren't updated, because there is no any case for RKE CIS and these images, right? thanks!

No. This was an accidental omission from my end. We should also make the same change in ubi images. Will update

helper_scripts/check_files_owner_in_dir.sh
afdesk: I meant maybe we can use `wget`/`curl` inside Dockerfile to download the script from sources.
my concern is we have to keep this one up to date manually now

afdesk (approved):

deven0t (approved):

PR #1739, opened 2024-11-28
FIX | k3s-cis-*- CHECK 4.2.1-4.2.3

aprilpear9696:
The check [4.2.1-4.2.3 for k3s-cis-*](https://github.com/aquasecurity/kube-bench/blob/main/cfg/k3s-cis-1.8/node.yaml#L155) is aim to check the node kubelet status. But current test case audit/audit_config commands are targeting kube-apiserver instead of kubelet.
Update the target tests' audit/audit_config command from checking kube-apiserver arguments to checking kubelet ones.


afdesk (approved):
@aprilpear9696 
 It seems it makes sense! 
thanks for the contribute!

afdesk:
@mozillazg wdyt?

mozillazg:
cc @andypitcher

mozillazg (approved):

afdesk:
merged
@aprilpear9696, @mozillazg thanks!

PR #1766, opened 2025-01-02
fix: CVE-2024-45338[CRITICAL]: Inefficient Regular Expression Complexity

jdesouza:
An attacker can craft an input to the Parse functions that would be processed non-linearly with respect to its length, resulting in extremely slow parsing. This could cause a denial of service.

https://avd.aquasec.com/nvd/2024/cve-2024-45338/


mozillazg (approved):

afdesk (approved):
LGTM!
thanks for the contribution

PR #1792, opened 2025-01-28
A collection of fixes for rke, fixes #1780

### LINKED ISSUE #1780: Required fixes for rke-cis 1.7 / 1.28 / 1.29
Following issues has to be updated 

- 1.1.11 => folder name should be updated from /node/var/lib/etcd to /var/lib/etcd
- 1.3.7 / 1.4.2  => the check is wrong - the reuqired logic is to find if bind-address is exists and equal to 127.0.0.1

   - id: 1.3.7

        text: "Ensure that the --bind-address argument is set to 127.0.0.1 (Automated)"
        audit: "/bin/ps -ef | grep $controllermanagerbin | grep -v grep"
        tests:
          bin_op: or
          test_items:
            - flag: "--bind-address"
              compare:
                op: eq
                value: "127.0.0.1"
            - flag: "--bind-address"
              set: false
        remediation: |
          Edit the Controller Manager pod specification file $controllermanagerconf
          on the control plane node and ensure the correct value for the --bind-address parameter
        scored: true

run example:

ubuntu@ip-10-0-5-7: /bin/ps -ef | grep kube-controller-manager | grep -v grep
root        2333    2313  1 Jan19 ?        00:14:07 kube-controller-manager --kubeconfig=/etc/kubernetes/ssl/kubecfg-kube-controller-manager.yaml --v=2 --cloud-provider= --service-cluster-ip-range=10.43.0.0/16 --terminated-pod-gc-threshold=1000 --cluster-cidr=10.42.0.0/16 --enable-hostpath-provisioner=false --allocate-node-cidrs=true --allow-untagged-cloud=true --configure-cloud-routes=false --leader-elect=true --profiling=false --authentication-kubeconfig=/etc/kubernetes/ssl/kubecfg-kube-controller-manager.yaml --service-account-private-key-file=/etc/kubernetes/ssl/kube-service-account-token-key.pem --authorization-kubeconfig=/etc/kubernetes/ssl/kubecfg-kube-controller-manager.yaml --node-monitor-grace-period=40s --pod-eviction-timeout=5m0s --root-ca-file=/etc/kubernetes/ssl/kube-ca.pem --use-service-account-credentials=true
ubuntu@ip-10-0-5-7:

but the expected result shows:

Expected Result
'--bind-address' is present OR '--bind-address' is not present

- 4.2.8  should be changed to skipped test since it is a manual test => https://ranchermanager.docs.rancher.com/reference-guides/rancher-security/hardening-guides/rke2-hardening-guide/rke2-self-assessment-guide-with-cis-v1.7-k8s-v1.25-v1.26-v1.27#428-ensure-that-the-eventrecordqps-argument-is-set-to-a-level-which-ensures-appropriate-event-capture-manual
- 4.2.12  should be changed to skipped test since it is a manual test => https://ranchermanager.docs.rancher.com/reference-guides/rancher-security/hardening-guides/rke2-hardening-guide/rke2-self-assessment-guide-with-cis-v1.7-k8s-v1.25-v1.26-v1.27#4212-ensure-that-the-kubelet-only-makes-use-of-strong-cryptographic-ciphers-manual

ttousai:


guyjerby (commented):
Hi @ttousai ,
it seems that there is no update of check 4.2.8 for rke-cis-1.23 to skip, can you check?
in addition 1.3.7 - still shows wrong logic - it should be failed since no 127.0.0.1 but it passed

guyjerby (approved):

afdesk (approved):

PR #1799, opened 2025-02-06
Doc bugfix. Updated version in installation documentation and using a version var

gekart:
The documentation used 0.6.2, a version that doesn't even install... This is a quick fix that uses a currently latest working version (0.10.1). It also combines all version references into a var. Beautify to your liking.

TODO: automate the version in release pipeline.


afdesk (approved):
@gekart thanks for the contribution!

PR #1841, opened 2025-03-21
add necessary mounts for /var/vcap/data/jobs & sys

bnutzinger:
Running kube-bench on TKGI with the provided job template will fail with errors like:
`14:55:43.616764 3064074 check.go:190] failed to run: "grep  ^tlsCertFile:\\s\\\"\\/var\\/vcap\\/jobs\\/kubelet\\/config\\/kubelet\\.pem\\\"\\ntlsPrivateKeyFile:\\s\\\"\\/var\\/vcap\\/jobs\\/kubelet\\/config\\/kubelet-key\\.pem\\\"$\n/var/vcap/jobs/kubelet/config/kubeletconfig.yml", output: "/bin/sh: /var/vcap/jobs/kubelet/config/kubeletconfig.yml: not found\n", error: exit status 127`

This is due to the fact that /var/vcap/jobs/<subdir> only contains symbolic links to /var/vcap/data/jobs/<subdir>/<uuid>
To actually read - for example - the kubeletconfig.yml the /var/vcap/data/jobs directory needs to be mounted aswell.

Likewise /var/vcap/sys is a symlink pointing to /var/vcap/data/sys

### DIFF
--- a/job-tkgi.yaml
+++ b/job-tkgi.yaml
@@ -23,6 +23,9 @@ spec:
             - name: var-vcap-jobs
               mountPath: /var/vcap/jobs
               readOnly: true
+            - name: var-vcap-data-jobs
+              mountPath: /var/vcap/data/jobs
+              readOnly: true
             - name: var-vcap-packages
               mountPath: /var/vcap/packages
               readOnly: true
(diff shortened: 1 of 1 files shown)


mozillazg (approved):

PR #1874, opened 2025-05-14
Add AKS-1.7 version

LaibaBareera:
Adding AKS 1.7.0 benchmark
[CIS Azure Kubernetes Service (AKS) Benchmark v1.7.0 PDF.pdf](https://github.com/user-attachments/files/20455539/CIS.Azure.Kubernetes.Service.AKS.Benchmark.v1.7.0.PDF.pdf)


cmd/util.go
```diff
@@ -490,6 +490,8 @@ func getPlatformBenchmarkVersion(platform Platform) string {
 	switch platform.Name {
 	case "eks":
 		return "eks-1.5.0"
+	case "aks":
```
mjshastha, line 493: Most managed clusters embed the platform name in the Kubernetes version (e.g., "v1.32.3-gke.1785003"), but AKS doesn't. So, our current logic fails to detect AKS clusters—we’ll need an alternative method to identify them and apply AKS-specific CIS rules.

cmd/util.go
```diff
@@ -485,11 +494,33 @@ func getPlatformInfoFromVersion(s string) Platform {
 	}
 }
 
+func IsAKS(ctx context.Context, k8sClient kubernetes.Interface) (bool, error) {
+	// Query the nodes for any annotations that indicate AKS (Azure Kubernetes Service)
```
mjshastha, line 498: The IsAKS function relies on the azure-identity-binding annotation, which isn't present in all AKS clusters. Instead, check reliable AKS indicators like labels are any other.

cfg/aks-1.7/node.yaml
```diff
@@ -0,0 +1,283 @@
...
+            - flag: root:root
+        remediation: |
+          Run the below command (based on the file location on your system) on the each worker node.
+          For example,
+          chown root:root $kubeletkubeconfig
+        scored: false
+
+      - id: 3.1.3
+        text: "Ensure that the azure.json file has permissions set to 644 or more restrictive (Automated)"
+        audit: '/bin/sh -c ''if test -e $kubeletconf; then stat -c permissions=%a $kubeletconf; fi'' '
```
guyjerby: @LaibaBareera - it should validate the azure.json file

cfg/aks-1.7/policies.yaml
```diff
@@ -0,0 +1,349 @@
...
+        scored: false
+
+
+  - id: 4.2
+    text: "Pod Security Policies"
+    checks:
+      - id: 4.2.1
+        text: "Minimize the admission of privileged containers (Automated)"
+        audit: |
+          kubectl get pods --all-namespaces -o json | jq -r '.items[] | select(.spec.containers[].securityContext.privileged == true) | .metadata.name'
```
guyjerby: @LaibaBareera the command retrieves list of objects that already filtered with privileged == true , so we just see if we have any response instead of searching the privileged  value

afdesk (approved):

guyjerby (changes requested):
Hi Laiba,

a few comments regarding the benchmark:

for 3.1 checks - scored should be true based on the definition from Manoj (automated test: scored = true ; manual test: scored = false)
for 4.1 checks - scored should be true based on the definition from Manoj (automated test: scored = true ; manual test: scored = false)
4.4.2 - missing automated test
4.5.1 - missing automated test
4.6.3 - missing automated test
5.1.1 - missing automated test
5.4.1 - 5.4.4 - missing automated test

guyjerby (approved):

afdesk:
@LaibaBareera could you fix the yaml linter errors pls?

PR #1945, opened 2025-09-05
add checks for cis benchmarks of rh-1.8

LaibaBareera:
- Added CIS Benchmark for Red Hat OpenShift Container Platform (OCP 4.13+)
Reference: [CIS_Red_Hat_Openshift_Container_Platform_Benchmark_v1.8.0_PDF.pdf](https://github.com/user-attachments/files/22170886/CIS_Red_Hat_Openshift_Container_Platform_Benchmark_v1.8.0_PDF.pdf)

- Implemented automation for benchmark selection

   - Run rh-1.4 when OCP version is 4.11 or 4.12

   - Run rh-1.8 when OCP version is above 4.12


cmd/util.go
```diff
@@ -623,7 +625,7 @@ func getOpenShiftInfo() Platform {
 
 func getOcpValidVersion(ocpVer string) (string, error) {
 	ocpOriginal := ocpVer
-	valid := []string{"3.10", "4.1", "4.11", "4.12", "4.13"}
+	valid := []string{"3.10", "4.1", "4.11", "4.13"}
```
afdesk, line 628: we don't support v4.12 anymore. is it intended?

LaibaBareera: Hey @afdesk , it supports OCP v4.12. For this version, it runs cis benchmark of rh-1.4. In this function, we’re defining the minimum supported OCP versions for each CIS benchmark release. For example, rh-1.0 supports OCP 4.1 through 4.10, rh-1.4 supports OCP 4.11 and 4.12, and rh-1.8 supports OCP 4.13 and above.

Issue #1953, opened 2025-09-11
Configuration for EKS v1.7.0 does not allow default value for eventRecordQPS


markusboehme:
**Overview**

The configuration of controls for the Amazon EKS CIS Benchmark v1.7.0 does not allow the default value for eventRecordQPS. This is recommendation 3.2.7 "Ensure that the --eventRecordQPS argument is set to 0 or a level which ensures appropriate event capture (Automated)".

**How did you run kube-bench?**

1. Modified `job-eks.yaml` to use `--benchmark 1.7.0`.
2. `kubectl apply -f job-eks.yaml`

**What happened?**

```
$ kubectl logs job/kube-bench
[...]
[FAIL] 3.2.7 Ensure that the --eventRecordQPS argument is set to 0 or a level which ensures appropriate event capture (Automated)
[...]
```

**What did you expect to happen:**

On manual inspection, I expected control 3.2.7 to pass.

**Environment** 

- kube-bench v0.12.0
- Amazon EKS v1.33.3

**Running processes**

`--event-qps` is not passed on the command line:

```
$ ps -eaf | grep kube 
root         582       1  0 Sep09 ?        00:19:45 /usr/bin/kubelet --hostname-override=ip-redacted.ec2.internal --config=/etc/kubernetes/kubelet/config.json --config-dir=/etc/kubernetes/kubelet/config.json.d --kubeconfig=/var/lib/kubelet/kubeconfig --image-credential-provider-bin-dir=/etc/eks/image-credential-provider --image-credential-provider-config=/etc/eks/image-credential-provider/config.json --node-ip=redacted --cloud-provider=external --node-labels=nodegroup=redacted-ng,alpha.eksctl.io/cluster-name=redacted,alpha.eksctl.io/nodegroup-name=redacted-ng
root         828     667  0 Sep09 ?        00:00:18 kube-proxy --v=2 --config=/var/lib/kube-proxy-config/config --hostname-override=ip-redacted.ec2.internal
ssm-user  691643  691565  0 08:48 pts/0    00:00:00 grep kube
```

**Configuration files**

`eventRecordQPS` is not defined in the configuration files:

```
$ sudo grep -R eventRecordQPS /etc/kubernetes/*
$ 
```

**Anything else you would like to add:**

- There has been discussion of this recommendation in the CIS benchmark for Amazon EKS before: https://github.com/aquasecurity/kube-bench/issues/1104#issuecomment-1250041648 
[...]

PR #1954, opened 2025-09-11
eks-1.7.0: allow default value for eventRecordQPS rule

markusboehme:
Closes: #1953 

The CIS Benchmark for Amazon EKS v1.7.0, recommendation 3.2.7 asks to "Ensure that the --eventRecordQPS argument is set to 0 or a level which ensures appropriate event capture". The --event-qps option on the command line and the eventRecordQPS option in the configuration file both have the same default value of 5, but differ in how they treat the an explicitly set value of 0:

  - The --event-qps command line option treats 0 as the default value of 5 QPS.
  - The eventRecordQPS configuration file option treats 0 as unlimited (and the absence of the option as the default value of 5 QPS).

Since setting --event-qps=0, using the default value, is acceptable for the command line option, using the default value for eventRecordQPS by not explicitly setting the option should be allowed as well. Note that this is already the case in the configuration for the generic Kubernetes CIS Benchmark.

With neither `--event-qps` passed on the command line nor `eventRecordQPS` defined in the configuration file, before:

```
[FAIL] 3.2.7 Ensure that the --eventRecordQPS argument is set to 0 or a level which ensures appropriate event capture (Automated)
```

After:

```
[PASS] 3.2.7 Ensure that the --eventRecordQPS argument is set to 0 or a level which ensures appropriate event capture (Automated)
```


afdesk:
@LaibaBareera it looks good to me, wdyt?

afdesk:
@markusboehme thanks for your contribution! LGTM

markusboehme:
Thanks for the review, @afdesk! I force pushed to update my committer identity to align with the signed CLA, and pick up changes from `main`.

LaibaBareera:
LGTM too.

PR #1972, opened 2025-10-02
Update golangci lint to v2

tylerauerbeck:
This updates the golangci-lint config to use the new v2 schema. It also bumps the version of golangci-lint used in the build workflow to use the latest v2 release.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,10 +1,28 @@
----
+version: "2"
 linters:
-  disable-all: true
+  default: none
   enable:
     - gocyclo
-    - gofmt
-    - goimports
     - govet
     - misspell
-    - typecheck
+  exclusions:
+    generated: lax
+    presets:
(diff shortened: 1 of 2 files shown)


tylerauerbeck:
Looks like the action itself needed bumped as well.

afdesk:
@tylerauerbeck thanks for your contribution

@mozillazg @LaibaBareera WDYT about this update? thanks!

.golangci.yaml
```diff
@@ -1,10 +1,28 @@
...
+      - legacy
+      - std-error-handling
+    paths:
+      - third_party$
+      - builtin$
+      - examples$
+formatters:
+  enable:
+    - gofmt
+    - goimports
```
afdesk, line 22: maybe `gocritic` should be added?

LaibaBareera:
LGTM too.

PR #1977, opened 2025-10-15
Add k3s-cis-1.11 benchmark configuration

jbenzel:
## Summary
Adds CIS Kubernetes Benchmark v1.11 configuration for K3s clusters (versions 1.29-1.34).

## Changes
- Added `cfg/k3s-cis-1.11/` directory with complete benchmark configuration
  - config.yaml (component settings)
  - master.yaml (60 control plane checks - Sections 1.1-1.4)
  - etcd.yaml (7 etcd checks - Section 2)
  - controlplane.yaml (3 control plane config checks - Section 3)
  - node.yaml (25 worker node checks - Sections 4.1-4.3)
  - policies.yaml (35 policy checks - Sections 5.1-5.6)
- Updated `cfg/config.yaml` with k3s-cis-1.11 version and target mappings


afdesk:
@jbenzel thanks for your contribution! it's really nice!

afdesk:
@LaibaBareera could you pls take a look too when you have time?

afdesk:
@jbenzel thanks!
could you sign `license/cla`?

andypitcher:
ref: https://github.com/rancher/security-scan/tree/main/package/cfg/k3s-cis-1.11

cc @dereknola

cfg/k3s-cis-1.11/master.yaml
```diff
@@ -0,0 +1,975 @@
...
+                  op: bitmask
+                  value: "600"
+          remediation: |-
+            Not Applicable.
+            By default, K3s embeds the controller manager within the k3s process. There is no controller manager pod specification file.
+          scored: true
+        - id: 1.1.4
+          text: Ensure that the controller manager pod specification file ownership is set to root:root (Automated)
+          type: skip
+          audit: /bin/sh -c 'if test -e $controllermanagerconf; then stat -c %U:%G $controllermanagerconf; fi'
```
LaibaBareera, line 53: @jbenzel, I noticed this check is marked as "skipped". Could you clarify why? Instead of skipping it, you could set the type to "manual".

cfg/k3s-cis-1.11/policies.yaml
```diff
@@ -0,0 +1,576 @@
...
+            else
+              printf "**clusterrole_name: %-50s is_whitelist: false is_compliant: true\n" "${clusterrole_name}"
+            fi;
+          done
+        tests:
+          test_items:
+            - flag: "is_compliant"
+              compare:
+                op: eq
+                value: "true"
```
LaibaBareera, line 85: If you set the type to "manual", you don’t need to include the test— it can be skipped automatically.

LaibaBareera (commented):
Thanks for your contribution.

andypitcher:
@jbenzel @benzeljd  I have a quick question about the origin of your PR (being one of the maintainer of the k3s/rke2/rke official profiles)

Was this created entirely independently, or was any of it adapted from existing code using the recent [security-scan project k3s-cis-1.11](https://github.com/rancher/security-scan/tree/main/package/cfg/k3s-cis-1.11) ? If it utilizes or is based on some prior work, please ensure a clear reference is added (e.g., in a comment or the PR description) to align with attribution requirements of its license. 

Thanks for your contribution !

cfg/k3s-cis-1.11/node.yaml
```diff
@@ -0,0 +1,459 @@
...
+            - flag: root:root
+        remediation: |
+          Not Applicable.
+          The kubelet is embedded in the k3s process. There is no kubelet config file, all configuration is passed in as arguments at runtime.
+        scored: true
+
+  - id: 4.2
+    text: "Kubelet"
+    checks:
+      - id: 4.2.1
```
ShashankSanil, line 152: @jbenzel  The current audit logic only checks the latest Running kubelet entry from journalctl, and if that log does not contain the --anonymous-auth flag, the check immediately falls back to the default value.

This is problematic for k3s because a kubelet flag may not always appear in the last log entry, even though it is still configured correctly in the kubelet config file ($kubeletconf). In such cases, kubebench will produce a false failure.

To avoid this, the audit logic should behave as follows:

If logs are present and the flag is found in the log, use that value.

If logs are present but the flag is not found, fall back to checking the kubelet config file.

If logs are not present at all, also check the config file.

Only if both sources are missing should we assume the default value.

This matches how other kubelet checks in kubebench behave and ensures correctness for k3s environments where kubelet args are sometimes set in config rather than in runtime logs.

I recommend updating the audit command accordingly to avoid false negatives.

andypitcher: FYI a [feature](https://github.com/k3s-io/k3s/issues/13223) is being worked on to avoid relying on `journalctl` but rather having a dedicated log file overwritten with the running components process details, each time k3s server is restarted/reloaded.

PR #2100, opened 2026-05-14
add benchmark for gke-1.9

LaibaBareera:


cmd/util.go
```diff
@@ -573,10 +573,12 @@ func gkeBenchmark(version string) string {
...
-	case "1.28", "1.29", "1.30":
+	case "1.28", "1.29":
 		return "gke-1.6.0"
-	case "1.31", "1.32", "1.33", "1.34":
+	case "1.30":
 		return "gke-1.8.0"
+	case "1.31", "1.32", "1.33", "1.34":
+		return "gke-1.9.0"
 	default:
 		return "gke-1.2.0"
```
eyalke: Suggest default to be the latest we have which is: gke-1.9.0

eyalke (approved):

PR #2103, opened 2026-05-20
Add cis benchmark for rh-1.9

LaibaBareera:


cfg/rh-1.9/node.yaml
```diff
@@ -0,0 +1,617 @@
...
+      - id: 4.1.5
+        text: "Ensure that the --kubeconfig kubelet.conf file permissions are set to 644 or more restrictive (Automated)"
+        audit: |
+          for node in $(oc get nodes -o jsonpath='{.items[*].metadata.name}'); do
+            echo "Node: ${node}"
+            oc debug node/${node} -- chroot /host stat -c '%a %n' /etc/kubernetes/kubelet.conf
+          done
+        use_multiple_values: true
+        tests:
+          test_items:
```
afdesk: it seems this check doesn't work correctly.

```sh
$ stat -c '%a %n' /etc/kubernetes/kubelet.conf
644 /etc/kubernetes/kubelet.conf
```
There is no `permissions` here.

the same case for rf-1.8 has another option: https://github.com/aquasecurity/kube-bench/blob/a73ce378418c71c5861e6963753f50dec3badadf/cfg/rh-1.8/node.yaml#L91

cfg/rh-1.9/policies.yaml
```diff
@@ -0,0 +1,499 @@
...
+          Then apply appropriate RBAC to assign this SCC only to necessary service accounts, groups, or users.
+          Carefully avoid assigning `allowPrivilegedContainer: true` in any SCC that is broadly bound.
+        scored: false
+
+      - id: 5.2.2
+        text: "Minimize the admission of containers wishing to share the host process ID namespace (Manual)"
+        type: manual
+        audit: |
+          oc get scc -o json \
+            | jq -r '[.items[] | select(.allowHostPID==true) | .metadata.name]
```
afdesk: could you pls check that `select(.allowHostPID==true)` is correct? it seems we need use `select(.allowHostPID==false)` here.

cfg/rh-1.9/master.yaml
```diff
@@ -0,0 +1,1336 @@
...
+          # For CNI multus
+          # Get the pod name in the openshift-multus namespace
+          POD_NAME=$(oc get pods -n openshift-multus -l app=multus --field-selector spec.nodeName="$NODE_NAME" -o jsonpath='{.items[0].metadata.name}' 2>/dev/null)
+
+          if [ -z "$POD_NAME" ]; then
+           echo "No matching pods found on the current node."
+          else
+          # Execute the stat command
+          oc exec -n openshift-multus "$POD_NAME" -- /bin/bash -c "stat -c '$i %n %U:%G' /host/etc/cni/net.d/*.conf" 2>/dev/null
+          oc exec -n openshift-multus $i -- /bin/bash -c "stat -c '$i %n %U:%G' /host/var/run/multus/cni/net.d/*.conf"  2>/dev/null
```
afdesk: what is `$i` here? it seems there is an undefined variable here

cfg/rh-1.9/master.yaml
```diff
@@ -0,0 +1,1336 @@
...
+          In this file, choose aescbc, kms or secretbox as the encryption provider.
+        scored: false
+
+      - id: 1.2.30
+        text: "Ensure that the API Server only makes use of Strong Cryptographic Ciphers (Manual)"
+        type: manual
+        audit: |
+          oc get cm -n openshift-authentication v4-0-config-system-cliconfig -o jsonpath='{.data.v4\-0\-config\-system\-cliconfig}' | jq .servingInfo
+          oc get kubeapiservers.operator.openshift.io cluster -o json |jq.spec.observedConfig.servingInfo
+          oc get openshiftapiservers.operator.openshift.io cluster -o json |jq.spec.observedConfig.servingInfo
```
afdesk:
```suggestion
          oc get openshiftapiservers.operator.openshift.io cluster -o json | jq.spec.observedConfig.servingInfo
```

cfg/rh-1.9/master.yaml
```diff
@@ -0,0 +1,1336 @@
...
+        use_multiple_values: true
+        tests:
+          test_items:
+            - flag: "permissions"
+              compare:
+                op: bitmask
+                value: "600"
+        remediation: |
+          No remediation required; file permissions are managed by the operator.
+        scored: false
```
afdesk, line 36: @LaibaBareera could you pls clarify for me, why all checks are marked `scored: false` now? Thanks!

cfg/rh-1.9/master.yaml
```diff
@@ -0,0 +1,1336 @@
...
+          if [ -z "$POD_NAME" ]; then
+          echo "No matching pods found on the current node."
+          else
+          # Execute the stat command
+          oc exec -n openshift-etcd "$POD_NAME" -- stat -c "$POD_NAME %n %U:%G" /var/lib/etcd/member
+          fi
+        use_multiple_values: true
+        tests:
+          test_items:
+            - flag: "root:root"
```
afdesk: the title says about it should be set to `etcd:etcd`, but there is `root:root` here

cfg/rh-1.9/node.yaml
```diff
@@ -0,0 +1,617 @@
...
+            fi
+          done
+        use_multiple_values: true
+        tests:
+          test_items:
+            - flag: "mode=Webhook"
+        remediation: |
+          No remediation required. By default, OpenShift uses secure authorization modes such as 'Webhook' and does not allow AlwaysAllow.
+          If AlwaysAllow is found, the node must be reconfigured using a KubeletConfig applied through the appropriate MachineConfigPool.
+        scored: true
```
afdesk, line 367: it's interesting. this check has a different way from https://github.com/aquasecurity/kube-bench/blob/a73ce378418c71c5861e6963753f50dec3badadf/cfg/rh-1.8/node.yaml#L240

now it passes only for `mode=Webhook`, is it correct? 
i think it's a bit confusing `the --authorization-mode argument is not set to AlwaysAllow`

afdesk (changes requested):

cmd/util.go
```diff
@@ -596,6 +596,8 @@ func ocpBenchmark(version string) string {
 		return "rh-1.4"
 	case "4.13":
 		return "rh-1.8"
+	case "4.14":
+		return "rh-1.9"
 	default:
 		return ""
```
eyalke, line 602: Please have the default as "rh-1.9"

cfg/rh-1.9/master.yaml
LaibaBareera: These are marked scored: false because they are Manual CIS/OCP checks, so we shouldn’t enforce them as hard PASS/FAIL scoring gates.
I still added test logic where possible to surface expected conditions (e.g., permissions bitmask), but results can require review/context.
So they show up to the customer as a review/warning item without incorrectly failing the whole benchmark.

cmd/util.go
LaibaBareera: I considered using "rh-1.9" as the default, but the current implementation intentionally resolves the OCP version to the nearest supported version before selecting a benchmark.

For example, OCP 4.14+ maps to rh-1.9, OCP 4.13 maps to rh-1.8, OCP 4.11 maps to rh-1.4, etc. The version resolution logic ensures that a cluster always receives the benchmark associated with the closest supported OCP version.

If we set the default directly to rh-1.9, any unsupported or unmapped version would automatically receive rh-1.9, bypassing the version-mapping logic and potentially assigning a benchmark that is not intended for that OCP release.

cfg/rh-1.9/policies.yaml
LaibaBareera: I’ve updated the audit command for this check.
Since this is a Manual check, the results still need to be reviewed, so we keep it marked as manual.

cfg/rh-1.9/master.yaml
afdesk: oh, all checks are `Manual` for `master`, got it. thanks

afdesk (approved):

cmd/util.go
eyalke, line 602: Please add the default to be "rh-1.9"

eyalke (changes requested):

eyalke (approved):

PR #2111, opened 2026-06-04
Fix CIS 4.3.1 false positive when kube-proxy is not running

arpitjain099:
Fixes #2036

## Problem

CIS check 4.3.1 ("Ensure that the kube-proxy metrics service is bound to localhost") audits the running kube-proxy process with `/bin/ps -fC $proxybin`. On a node where no kube-proxy process is running, `ps` exits with status 1. `runAuditCommands()` returns that error before any test item is evaluated, so a scored check is forced straight to FAIL with a `failed to run: "/bin/ps -fC proxy" ... exit status 1` reason.

That is a false positive. If kube-proxy is not running, the metrics port (10249) is not listening at all, so it cannot be bound to a non-loopback address. The control is not applicable and should report PASS rather than FAIL. This was reported in #2036 on a node where the proxy component is provided by a different CNI and no `kube-proxy`/`proxy` process exists.

## Fix

Append `|| true` to the 4.3.1 audit so the process listing always exits 0 (it still prints its header-only output when nothing matches). With a successful exit the existing test runs, and because the check already uses

```yaml
tests:
  bin_op: or
  test_items:
    - flag: "--metrics-bind-address"
      ...
      compare: { op: has, value: "127.0.0.1" }
    - flag: "--metrics-bind-address"
      set: false
```

the second branch (`--metrics-bind-address` not present) evaluates to true, so the check PASSes when kube-proxy is absent.

When kube-proxy IS running, the audit output is unchanged, so the bind-address comparison still applies and a real non-localhost binding is still caught. No detection capability is lost.

`|| true` is the same idiom kube-bench already uses elsewhere in the configs to tolerate a non-zero exit while keeping command output.

The same audit line exists in cis-1.9, cis-1.10, cis-1.11 and cis-1.12, so the change is applied to all four for consistency. (k3s-cis-1.9 uses a different journalctl-based audit and is unaffected.)

## Before / after

Before (no kube-proxy process running):

```
[FAIL] 4.3.1 Ensure that the kube-proxy metrics service is bound to localhost (Automated)
  reason: failed to run: "/bin/ps -fC proxy", output: "UID PID PPID C STIME TTY TIME CMD\n", error: exit status 1
```

After:

```
[PASS] 4.3.1 Ensure that the kube-proxy metrics service is bound to localhost (Automated)
```

## Tests

Added a regression case to `TestCheck_Run` in `check/check_test.go` that models the process-absent scenario (scored check, `bin_op: or`, audit with no `--metrics-bind-address` flag in output) and asserts PASS. Confirmed the same shape FAILs without the tolerant exit, so the test guards the regression.

`make tests` passes.

### DIFF
--- a/cfg/cis-1.10/node.yaml
+++ b/cfg/cis-1.10/node.yaml
@@ -459,7 +459,7 @@ groups:
     checks:
       - id: 4.3.1
         text: "Ensure that the kube-proxy metrics service is bound to localhost (Automated)"
-        audit: "/bin/ps -fC $proxybin"
+        audit: "/bin/ps -fC $proxybin || true"
         audit_config: "/bin/sh -c 'if test -e $proxykubeconfig; then cat $proxykubeconfig; fi'"
         tests:
           bin_op: or
--- a/cfg/cis-1.11/node.yaml
+++ b/cfg/cis-1.11/node.yaml
@@ -490,7 +490,7 @@ groups:
     checks:
       - id: 4.3.1
         text: "Ensure that the kube-proxy metrics service is bound to localhost (Automated)"
-        audit: "/bin/ps -fC $proxybin"
+        audit: "/bin/ps -fC $proxybin || true"
         audit_config: "/bin/sh -c 'if test -e $proxykubeconfig; then cat $proxykubeconfig; fi'"
         tests:
           bin_op: or
--- a/cfg/cis-1.12/node.yaml
+++ b/cfg/cis-1.12/node.yaml
@@ -473,7 +473,7 @@ groups:
     checks:
       - id: 4.3.1
         text: "Ensure that the kube-proxy metrics service is bound to localhost (Automated)"
-        audit: "/bin/ps -fC $proxybin"
+        audit: "/bin/ps -fC $proxybin || true"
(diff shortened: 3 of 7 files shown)


EdwardCooke:
Now that cis 2.0 is in there, you may want to update your PR to include that one. It has the same problem.

arpitjain099:
Good catch, thanks. Done, and the branch is rebased onto current main so cis-2.0 is actually in the tree.

While I was in there I grepped the whole of `cfg/` for the unguarded form rather than just adding the one you named:

```
$ grep -rn 'audit: "/bin/ps -fC $proxybin"' cfg/
cfg/cis-2.0/node.yaml:476
cfg/rke2-cis-1.9/node.yaml:429
```

So `rke2-cis-1.9` has it too, and I missed that one in the first pass. I have included it, since on RKE2 kube-proxy usually runs inside the supervisor process rather than as its own `kube-proxy` process, which makes a node with no match the normal case there rather than an edge case. Happy to drop it back out if you would rather keep this PR to the upstream CIS profiles.

That grep now comes back clean for all six profiles. `go test ./check/... ./cmd/...` is still green.

willemm:
Shouldn't the same be done for 4.1.3 and 4.1.4 ?

PR #2120, opened 2026-07-06
fix(cfg): don't fail CNI file checks when host-local IPAM dir is absent

### LINKED ISSUE #2119: CNI file checks 1.1.9/1.1.10 false-FAIL on CNIs without host-local IPAM (absent /var/lib/cni/networks)
## Summary

On a node whose CNI does not use the host-local IPAM plugin (Cilium, Calico, ...), checks 1.1.9 and 1.1.10 report a FAIL that is unrelated to any real file permission or ownership problem. The audit inspects `/var/lib/cni/networks`, which does not exist under those CNIs, so it produces no output, and an empty `use_multiple_values` result is scored as a failure.

## Affected checks

- 1.1.9 "Ensure that the Container Network Interface file permissions are set to 600 or more restrictive"
- 1.1.10 "Ensure that the Container Network Interface file ownership is set to root:root"

Audit (k3s profiles):

```
find /var/lib/cni/networks -type f ! -name lock 2> /dev/null | xargs --no-run-if-empty stat -c permissions=%a
find /var/lib/cni/networks -type f 2> /dev/null | xargs --no-run-if-empty stat -c %U:%G
```

## Root cause

`/var/lib/cni/networks` is the state directory of the host-local IPAM plugin: one file per allocated pod IP. Only CNIs that use host-local IPAM create it — including the flannel backend that ships as the k3s default. CNIs that manage their own IPAM (Cilium, Calico, ...) never create it, so the directory is absent.

When the directory is absent, `find` prints nothing (the "No such file or directory" error is swallowed by `2> /dev/null`), `xargs --no-run-if-empty` runs `stat` zero times, and the audit output is empty. With `use_multiple_values: true` an empty output means zero values to test, which the runner scores as a failure instead of a pass.

## Impact differs by profile

The generic `cis-*` profiles keep these checks `scored: false`, so the empty result surfaces as WARN. Several distro profiles mark them `scored: true` (Automated), where the same empty result becomes a hard FAIL. So on, for example, a k3s cluster running Cilium, kube-bench reports 1.1.9 and 1.1.10 as FAIL even though there is nothing to check.

## Affected profiles

Active checks that hardcode `/var/lib/cni/networks` and can produce this false result: `k3s-cis-1.7`, `k3s-cis-1.8`, `k3s-cis-1.9` (several marked Automated), and `rke-cis-1.7/1.23/1.24`, `rke2-cis-1.7/1.8/1.23` (mostly Manual; `rke-cis-1.24` 1.1.10 is Automated). The generic `cis-*` profiles carry the same `/var/lib/cni/networks` fallback but keep the checks Manual (WARN).

Already handled: `k3s-cis-1.23` and `k3s-cis-1.24` mark these checks `type: skip`; `rke2-cis-1.24` 1.1.10 already guards with an `else echo "File not found"` branch plus `bin_op: or`.

## Reproduction

On a k3s node using Cilium:

```
$ sudo ls /var/lib/cni/networks
ls: cannot access '/var/lib/cni/networks': No such file or directory

$ sudo find /var/lib/cni/networks -type f ! -name lock 2>/dev/null | xargs --no-run-if-empty stat -c permissions=%a
      # (no output)

$ sudo kube-bench run --benchmark k3s-cis-1.9 --targets master | grep -E '1\.1\.(9|10) '
[FAIL] 1.1.9 Ensure that the Container Network Interface file permissions are set to 600 or more restrictive (Automated)
[FAIL] 1.1.10 Ensure that the Container Network Interface file ownership is set to root:root (Automated)
```

## Related

- #1669 — same FAIL symptom via the generic profile and the deprecated `--cni-conf-dir` kubelet flag; the `/var/lib/cni/networks` case is a distinct trigger.
- #867 / #877 — silenced the "No such file or directory" stderr, but did not change the empty-output scoring.
- #902 — a similar Calico false positive, closed as too CNI-specific.

## Proposed fix

Make the audit tolerate an absent or empty `/var/lib/cni/networks`: when no host-local IPAM files exist, emit a passing sentinel so the check reports PASS; when files do exist, evaluate their real permissions/ownership unchanged. This keeps the checks Automated and meaningful on flannel-based clusters while removing the false positive on other CNIs. I have a patch ready.

lexfrei:
## What

Checks 1.1.9 and 1.1.10 audit CNI file permissions/ownership by running `find /var/lib/cni/networks`. That directory is the host-local IPAM state store, populated only by CNIs that use host-local IPAM (e.g. the flannel backend that ships as the k3s default). Under a CNI with its own IPAM (Cilium, Calico, ...) the directory does not exist, `find` returns nothing, and an empty `use_multiple_values` audit is scored as FAIL (Automated profiles) or WARN (Manual profiles) — a false positive unrelated to any real permission or ownership problem. See #2119.

## Change

Rewrite the 1.1.9/1.1.10 audit so an absent or empty `/var/lib/cni/networks` yields a passing sentinel value (`permissions=<mode>` / `root:root`) instead of an empty result. When host-local IPAM files do exist, their real permissions and ownership are still evaluated, so genuine issues continue to fail. The host-local `lock` file (created 0750) is also excluded from the permission check so it no longer trips the 600 bitmask (previously only 1.1.9 in the k3s profiles excluded it).

Profiles touched (both 1.1.9 and 1.1.10):

- `k3s-cis-1.7`, `k3s-cis-1.8`, `k3s-cis-1.9` (single-command audit)
- `rke-cis-1.7`, `rke-cis-1.23`, `rke-cis-1.24`
- `rke2-cis-1.7`, `rke2-cis-1.8`, `rke2-cis-1.23` (two-command audit)

Not touched, by design:

- `k3s-cis-1.23` / `k3s-cis-1.24` already mark these checks `type: skip`.
- `rke2-cis-1.24` 1.1.10 already guards with an `else echo "File not found"` branch and `bin_op: or`.
- The generic `cis-*` profiles keep these checks Manual (WARN) and stay aligned with upstream CIS; I left them as-is to avoid changing the CIS baseline. Happy to extend the same tolerance there if you prefer uniform coverage.

## Validation

Audit logic tested in isolation for both audit shapes: absent directory → passing sentinel; a real 600 file → `permissions=600`; a real 644 file → `permissions=644` (the failing value still surfaces); a 0750 `lock` file → excluded.

End-to-end on a k3s node running Cilium (`/var/lib/cni/networks` absent), same binary, config-dir difference only:

```
# before (released cfg)
[FAIL] 1.1.9  ...
[FAIL] 1.1.10 ...
# after (this change)
[PASS] 1.1.9  ...
[PASS] 1.1.10 ...
```

## Alternative considered

Aligning the k3s/rke/rke2 profiles to `scored: false` (WARN), matching the generic `cis-*` profiles, would hide the hard FAIL but also stop flagging real permission/ownership problems on clusters that do use host-local IPAM. The audit-tolerance approach keeps the checks Automated and meaningful for those clusters while removing the false positive on the others, so I chose it over down-scoring. A third option — the `type: skip` used by `k3s-cis-1.23/1.24` — disables the check even where it is valid; audit-tolerance is a strict improvement over that and could replace those skips in a follow-up.

Fixes #2119

### DIFF
--- a/cfg/k3s-cis-1.7/master.yaml
+++ b/cfg/k3s-cis-1.7/master.yaml
@@ -120,7 +120,12 @@ groups:
 
       - id: 1.1.9
         text: "Ensure that the Container Network Interface file permissions are set to 600 or more restrictive (Manual)"
-        audit: find /var/lib/cni/networks -type f ! -name lock 2> /dev/null | xargs --no-run-if-empty stat -c permissions=%a
+        audit: |
+          if [ -d /var/lib/cni/networks ] && [ -n "$(find /var/lib/cni/networks -type f ! -name lock 2>/dev/null)" ]; then
+            find /var/lib/cni/networks -type f ! -name lock 2>/dev/null | xargs --no-run-if-empty stat -c permissions=%a
+          else
+            echo "permissions=600"
+          fi
         use_multiple_values: true
         tests:
           test_items:
@@ -137,7 +142,12 @@ groups:
 
       - id: 1.1.10
         text: "Ensure that the Container Network Interface file ownership is set to root:root (Automated)"
-        audit: find /var/lib/cni/networks -type f 2> /dev/null | xargs --no-run-if-empty stat -c %U:%G
+        audit: |
+          if [ -d /var/lib/cni/networks ] && [ -n "$(find /var/lib/cni/networks -type f ! -name lock 2>/dev/null)" ]; then
+            find /var/lib/cni/networks -type f ! -name lock 2>/dev/null | xargs --no-run-if-empty stat -c %U:%G
+          else
+            echo "root:root"
+          fi
         use_multiple_values: true
         tests:
           test_items:
--- a/cfg/k3s-cis-1.8/master.yaml
+++ b/cfg/k3s-cis-1.8/master.yaml
@@ -120,7 +120,12 @@ groups:
 
       - id: 1.1.9
         text: "Ensure that the Container Network Interface file permissions are set to 600 or more restrictive (Automated)"
-        audit: find /var/lib/cni/networks -type f ! -name lock 2> /dev/null | xargs --no-run-if-empty stat -c permissions=%a
+        audit: |
+          if [ -d /var/lib/cni/networks ] && [ -n "$(find /var/lib/cni/networks -type f ! -name lock 2>/dev/null)" ]; then
+            find /var/lib/cni/networks -type f ! -name lock 2>/dev/null | xargs --no-run-if-empty stat -c permissions=%a
+          else
+            echo "permissions=600"
+          fi
         use_multiple_values: true
         tests:
           test_items:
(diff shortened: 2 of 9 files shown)


lexfrei:
Rebased onto current main, no conflicts. `make tests` passes locally.

CI here hasn't run at all, the workflow is sitting in `action_required` waiting on approval for a fork PR. Could you approve the run?

